PR #19, opened 2026-05-21
fix(auth): honor PRISMA_API_TOKEN in readAuthState for CI/headless use

### LINKED ISSUE #18: PRISMA_API_TOKEN is half-supported: app deploy fails in CI
## Summary

`PRISMA_API_TOKEN` is documented as the headless / CI auth surface (see [`guard.ts`](https://github.com/prisma/prisma-cli/blob/main/packages/cli/src/lib/auth/guard.ts) — "PRISMA_API_TOKEN env var → service token (CI / headless)"). But it is only honored for building the `ManagementApiClient`. **Workspace resolution still requires a local OAuth file**, so `app deploy`, `project list`, and every other command that goes through `requireAuthenticatedAuthState` is unusable headlessly.

## Repro

In a clean environment (no `~/.config/prisma/auth.json` / `~/Library/Application Support/prisma/auth.json`), with a valid workspace service token:

```bash
export PRISMA_API_TOKEN=<token where sub="workspace:<id>">
npx --package=@prisma/cli@preview --yes prisma-cli auth whoami --json
# → { "authenticated": false, ... }

npx --package=@prisma/cli@preview --yes prisma-cli app deploy --no-interactive
# → AUTH_REQUIRED
```

## Root cause

In [`lib/auth/auth-ops.ts:readAuthState`](https://github.com/prisma/prisma-cli/blob/main/packages/cli/src/lib/auth/auth-ops.ts), authentication state is read **exclusively from `FileTokenStorage`**:

```ts
const tokenStorage = new FileTokenStorage(env);
const tokens = await tokenStorage.getTokens();
if (!tokens) return { authenticated: false, ... };
```

`requireAuthenticatedAuthState` then sees `authenticated: false`, can't prompt in non-interactive mode, and throws `AUTH_REQUIRED` — even though `PRISMA_API_TOKEN` is set and `requireComputeAuth` would have happily used it for API calls.

Locally the bug also manifests as **deploying to the wrong workspace**: if you have any stored OAuth session, that workspaceId wins over the `sub` claim of `PRISMA_API_TOKEN`. We saw this with a token scoped to workspace A and a stored login for workspace B — deploy went to B, created an orphan project, and ignored the project pinned in our config.

## Proposed fix

When `PRISMA_API_TOKEN` is set, derive auth state from the token itself and skip `FileTokenStorage` entirely:

1. Decode the JWT.
2. Parse the workspace from `sub` (`"workspace:<id>"` → `<id>`).
3. Build a synthetic `AuthStateResult` with `authenticated: true` and the workspace populated. Hydrate `workspace.name` from `GET /v1/workspaces/{id}` when available; fall back to the id (matches the existing OAuth-path behavior).
4. If the token has no `workspace:` `sub`, return signed-out state so commands surface the standard `AUTH_REQUIRED` error instead of crashing.

This mirrors `requireComputeAuth`'s precedence (`PRISMA_API_TOKEN` over file storage) and keeps `auth whoami` and `app deploy` reading the same source of truth.

## Industry precedent

Vercel CLI's `VERCEL_TOKEN`, GitHub CLI's `GH_TOKEN`, Fly.io's `FLY_API_TOKEN`, Doppler's `DOPPLER_TOKEN`, Stripe's `STRIPE_API_KEY` — all behave this way: when the token env var is set, it is fully sufficient for headless operation and any locally cached credentials are ignored. The Prisma JWT already encodes the workspace in `sub`, so no additional `PRISMA_WORKSPACE_ID` env var is needed for the common case.

## Out of scope (follow-up)

`auth login` / `auth logout` still target `FileTokenStorage` when `PRISMA_API_TOKEN` is set, which is confusing. Worth a small follow-up: either short-circuit `auth login` to return the env-derived state with a warning, or refuse with a clear structured error.

## Impact

This currently blocks every CI deploy of `@pdp/control-plane` services using `prisma-cli app deploy` since `alpha.4`. We have a brittle workaround (seed a fake `auth.json` in CI) but the contract documented on `requireComputeAuth` should just work.

kristof-siket:
Closes #18.

## Summary

`PRISMA_API_TOKEN` is documented on [`requireComputeAuth`](https://github.com/prisma/prisma-cli/blob/main/packages/cli/src/lib/auth/guard.ts) as the headless / CI auth surface, but `readAuthState` reads exclusively from `FileTokenStorage`. The token is therefore sufficient to build a `ManagementApiClient` but not to satisfy `requireAuthenticatedAuthState`. Two symptoms follow:

- On a clean CI runner, `app deploy` (and every other command that resolves project context) fails with `AUTH_REQUIRED` even with a valid `PRISMA_API_TOKEN` set, because no OAuth file exists.
- On a developer machine with any stored OAuth session, that workspace wins over the token's `sub` claim, so deploys can silently target the wrong workspace. We hit this in the wild: a token scoped to workspace A, a stale login for workspace B, and `app deploy` created a new orphan project under B.

This PR makes `readAuthState` honor `PRISMA_API_TOKEN` the same way `requireComputeAuth` already does:

1. If `PRISMA_API_TOKEN` is set, decode the JWT, parse the workspace id from `sub` (`workspace:<id>`), and build the synthetic `AuthStateResult` from there — skipping `FileTokenStorage` entirely.
2. Hydrate `workspace.name` from `GET /v1/workspaces/{id}` when the API is reachable, falling back to the id on failure (matches existing OAuth-path behavior).
3. If the token has no `workspace:` subject, return signed-out state so callers produce the standard `AUTH_REQUIRED` error rather than crashing or deploying somewhere unintended.

The OAuth-file path is unchanged when `PRISMA_API_TOKEN` is unset or empty.

## Behavior comparison

With `PRISMA_API_TOKEN=<token scoped to workspace clitq5...>` and a locally stored OAuth login for an unrelated workspace `cmoijyfgr...`:

| CLI | `auth whoami --json` workspace |
| --- | --- |
| `@prisma/cli@preview` (alpha.4) | `cmoijyfgr0wlm3ucagwc75f3f` (the local OAuth file wins) |
| This PR | `clitq5hfg0000qv0gtg9nv9fy` (the service token wins) |

## Docs

Adds an Authentication section to `docs/product/command-spec.md` documenting the two auth sources and their precedence, since the document-driven development rule requires product docs to be the source of truth.

## Industry precedent

Standard CLI behavior: when the token env var is set, it is fully sufficient for headless operation and locally cached credentials are ignored.

- Vercel: `VERCEL_TOKEN` (plus `VERCEL_ORG_ID` / `VERCEL_PROJECT_ID` for explicit scope)
- GitHub CLI: `GH_TOKEN`
- Fly.io: `FLY_API_TOKEN`
- Doppler: `DOPPLER_TOKEN`
- Stripe: `STRIPE_API_KEY`

The Prisma JWT already encodes the workspace in `sub`, so no `PRISMA_WORKSPACE_ID` env var is needed for the common case where the token is scoped to a single workspace.

## Tests

- `derives authenticated state from PRISMA_API_TOKEN without consulting FileTokenStorage`
- `ignores a stored OAuth session when PRISMA_API_TOKEN is set` (regression test for the wrong-workspace scenario)
- `falls back to the workspace id when the API lookup fails for a service token`
- `returns signed-out state when PRISMA_API_TOKEN does not carry a workspace subject`
- `treats an empty PRISMA_API_TOKEN as unset and falls back to FileTokenStorage`

Full suite (`pnpm test`): 192 passed.

## Manual verification

Built locally and ran against the platform with a real service token:

```bash
# token sub: workspace:clitq5hfg0000qv0gtg9nv9fy
$ PRISMA_API_TOKEN=$TOKEN node packages/cli/dist/cli.js auth whoami --json
{
  "ok": true,
  "command": "auth.whoami",
  "result": {
    "authenticated": true,
    "provider": null,
    "user": null,
    "workspace": { "id": "clitq5hfg0000qv0gtg9nv9fy", "name": "clitq5hfg0000qv0gtg9nv9fy" }
  },
  "warnings": [],
  "nextSteps": []
}
```

Same invocation against `@prisma/cli@preview` (alpha.4) reports the local OAuth file's workspace, confirming the regression and the fix.

## Out of scope (follow-up)

`auth login` and `auth logout` still operate on `FileTokenStorage` regardless of `PRISMA_API_TOKEN`. That is documented in the new spec section but the UX could be tightened (e.g. `auth login` short-circuits to the env-derived state with a warning when the env var is set). Happy to do that as a separate PR if you'd like.

### DIFF
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -50,12 +50,12 @@ Out of scope for the current preview:
 
 The CLI accepts two authentication sources, in this fixed precedence:
 
-1. `PRISMA_API_TOKEN` environment variable — long-lived service token, intended for CI and other headless contexts.
+1. `PRISMA_SERVICE_TOKEN` environment variable — long-lived service token, intended for CI and other headless contexts.
 2. Stored OAuth session — created by `prisma-cli auth login`, kept in the OS-appropriate credentials store, refreshed automatically.
 
-When `PRISMA_API_TOKEN` is set, the token is fully sufficient for authenticated commands. The CLI does not read any locally stored OAuth session in that case, so behavior is identical on a fresh runner and a developer machine that happens to be signed in. The active workspace is derived from the token's `sub` claim; no additional flag or environment variable is required for the common case where the token is scoped to a single workspace.
+When `PRISMA_SERVICE_TOKEN` is set, the token is fully sufficient for authenticated commands. The CLI does not read any locally stored OAuth session in that case, so behavior is identical on a fresh runner and a developer machine that happens to be signed in. The active workspace is derived from the token's `sub` claim; no additional flag or environment variable is required for the common case where the token is scoped to a single workspace.
 
-`auth login` and `auth logout` operate on the stored OAuth session. They do not affect the `PRISMA_API_TOKEN` environment variable.
+`auth login` and `auth logout` operate on the stored OAuth session. They do not affect the `PRISMA_SERVICE_TOKEN` environment variable.
 
 ## Context Resolution
 
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -53,7 +53,7 @@ The CLI accepts two authentication sources, in this fixed precedence:
 1. `PRISMA_SERVICE_TOKEN` environment variable — long-lived service token, intended for CI and other headless contexts.
 2. Stored OAuth session — created by `prisma-cli auth login`, kept in the OS-appropriate credentials store, refreshed automatically.
 
-When `PRISMA_SERVICE_TOKEN` is set, the token is fully sufficient for authenticated commands. The CLI does not read any locally stored OAuth session in that case, so behavior is identical on a fresh runner and a developer machine that happens to be signed in. The active workspace is derived from the token's `sub` claim; no additional flag or environment variable is required for the common case where the token is scoped to a single workspace.
+When `PRISMA_SERVICE_TOKEN` is set and non-empty, the token is fully sufficient for authenticated commands. If `PRISMA_SERVICE_TOKEN` is set but empty or only whitespace, commands fail with an auth configuration error instead of falling back to stored OAuth. The CLI does not read any locally stored OAuth session when a non-empty service token is present, so behavior is identical on a fresh runner and a developer machine that happens to be signed in. The active workspace is derived from the token's `sub` claim; no additional flag or environment variable is required for the common case where the token is scoped to a single workspace.
 
 `auth login` and `auth logout` operate on the stored OAuth session. They do not affect the `PRISMA_SERVICE_TOKEN` environment variable.
 
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -46,6 +46,17 @@ Out of scope for the current preview:
 - Public Beta does not read or write `prisma.config.ts`, `.prisma/settings.json`, or any repo config file for Project -> Branch -> App resolution.
 - Remote commands do not silently change local context.
 
+## Authentication
+
+The CLI accepts two authentication sources, in this fixed precedence:
+
+1. `PRISMA_API_TOKEN` environment variable — long-lived service token, intended for CI and other headless contexts.
+2. Stored OAuth session — created by `prisma-cli auth login`, kept in the OS-appropriate credentials store, refreshed automatically.
+
+When `PRISMA_API_TOKEN` is set, the token is fully sufficient for authenticated commands. The CLI does not read any locally stored OAuth session in that case, so behavior is identical on a fresh runner and a developer machine that happens to be signed in. The active workspace is derived from the token's `sub` claim; no additional flag or environment variable is required for the common case where the token is scoped to a single workspace.
+
+`auth login` and `auth logout` operate on the stored OAuth session. They do not affect the `PRISMA_API_TOKEN` environment variable.
+
 ## Context Resolution
 
 ### Project
--- a/packages/cli/src/controllers/app.ts
+++ b/packages/cli/src/controllers/app.ts
@@ -1442,7 +1442,11 @@ async function resolveProjectContext(
     createProject: options?.allowCreate
       ? async (name) => {
           const project = await provider.createProject({ name }).catch((error) => {
-            throw deployFailedError("Failed to create project for first deploy", error, ["prisma-cli app deploy"]);
+            throw createProjectOnFirstDeployError({
+              error,
+              inferredName: name,
+              workspaceName: authState.workspace!.name,
+            });
           });
           return {
             id: project.id,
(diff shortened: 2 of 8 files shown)


kristof-siket:
## Follow-up commits: error UX papercuts from the same end-to-end test

While verifying the `PRISMA_API_TOKEN` precedence change on a real headless run, I hit three error-message papercuts that made a misconfigured token look fine, or made the resulting failure unhelpful. Each lands as its own commit on this branch so the diff stays reviewable.

### 1271c57 — `fix(auth): treat 401 on workspace lookup as signed out`

If `PRISMA_API_TOKEN` is set but the API rejects it with HTTP 401 (revoked / wrong signing key / expired), `auth whoami --json` was returning:

```json
{ "authenticated": true, "workspace": { "id": "<raw cuid>", "name": "<raw cuid>" } }
```

…because the `try/catch` in `buildAuthState` swallowed the 401 and fell through to the id-as-name placeholder. `buildAuthState` now distinguishes by response status:

- **401** → return signed-out auth state. `auth whoami` reports the truth; downstream commands trigger the normal `AUTH_REQUIRED` flow with a clear message.
- **404 / 5xx / network** → keep the existing silent fallback (id-as-name), since the token is presumably valid and the lookup is just transient.

Added two unit tests in `auth-ops.test.ts` covering both branches.

### 87e5bbb — `fix(app deploy): explain create-on-first-deploy failures clearly`

`app deploy` falls into "create a new project on first deploy" when no existing project matches the `package.json` name. When the create call was rejected, the user saw:

```json
{
  "error": {
    "code": "DEPLOY_FAILED",
    "summary": "Failed to create project for first deploy",
    "why": "Authentication failed (HTTP 401)",
    "fix": "Retry the command, or rerun with --trace for more detailed diagnostics."
  }
}
```

Two problems:
- The summary didn't reveal that the CLI was inferring projects from `package.json` and silently entering create-mode. The user expected the deploy to find an existing project.
- "Retry the command" is unhelpful for a 401 — retrying won't help.

The new error is context-aware:

- **401 / 403** → `AUTH_FORBIDDEN`
  - summary: `"Could not create a new project for this deploy"`
  - why: ``"No existing project matched the package.json name `<pkg>`, so the CLI attempted to create one. The platform rejected the create (HTTP 401)."``
  - fix: ``"Pass --project <id-or-name> to deploy into an existing project, or grant the service token project-create permission on workspace `<workspaceName>`."``
- **Other failures** → keep `DEPLOY_FAILED`, but use the new wording and append the underlying error message, and recommend `--project <id-or-name>` in nextSteps.

Added two unit tests in `app-controller.test.ts` covering both branches.

### 976de38 — `fix(project): suggest a concrete --project command on PROJECT_AMBIGUOUS`

`PROJECT_AMBIGUOUS` already includes `meta.matches` with id + name, but `nextSteps` only pointed at `prisma-cli project list`. Now we append `prisma-cli app deploy --project <first-match-id>` so the user can see the exact shape of the disambiguation flag in the matched-id format without guessing.

### Verification

- `env -u PRISMA_API_TOKEN pnpm test` — 196 passed (was 194 before; +2 from new app-controller cases; the +1 from auth-ops's pair was already absorbed into the file count).
- `pnpm build:cli` — succeeds.

luanvdw (commented):
Looks good overall!: PRISMA_API_TOKEN/headless auth should take precedence over cached OAuth state so CI and local machines behave consistently 🙏

Two small things before merging:
1. Naming: we currently phrase workspace-level access as “service tokens” in our product language, not “API tokens”. Since this PR adds/locks the documented headless auth contract, can we use `PRISMA_SERVICE_TOKEN` here?
2. Empty env var behavior: readAuthState treats an empty/whitespace token as unset and falls back to FileTokenStorage, while requireComputeAuth treats it as set-but-invalid and errors. That can make `auth whoami` report the cached OAuth workspace even though a later command fails because the env token is empty.

Otherwise the behavior change and tests look solid!

luanvdw (approved):
🚀

PR #32, opened 2026-05-25
Print auth URL to console to support a remote auth flow

sneub:
### Summary
`prisma-cli auth login` opens a browser and waits on a `localhost:<port>/auth/callback`
HTTP server for the OAuth redirect. This breaks on remote machines (SSH, Docker,
cloud VMs): no browser to open, and even if the user copies the URL to a local
browser, the redirect lands on the *local* machine's `localhost`, never reaching
the remote CLI.

This change adds a second completion path — paste-the-callback-URL — that runs in
parallel with the existing browser callback. No backend changes, no SDK changes,
no new flag. Laptop users see no behavior change; remote users get a path that
actually works.

### How it works
- The auth URL is now also printed to stderr with brief instructions.
- The local HTTP server still runs as before.
- A `readline` prompt (`Paste URL here:`) runs in parallel when stdin is a TTY.
- `Promise.race([httpCallback, pastedUrl])` — whichever delivers a callback URL
  first wins. Both paths call the same `state.handleCallback(url)`.
- A `completed` guard ensures the OAuth code exchange runs exactly once even if
  both fire close together; the losing path is aborted via `AbortController` and
  the existing server-close `finally`.
- `open(url)` failures are now swallowed instead of bubbling — the paste path
  is a viable fallback when the browser launch silently fails on headless boxes.

### UX notes
- On non-TTY stdin (CI, piped, tests) the paste prompt is skipped entirely;
  behavior is identical to today.
- Accepts the full callback URL (not just the `code`) so we can keep `state`
  validation intact with zero new parsing logic.

### Scope
Single file: `packages/cli/src/lib/auth/login.ts` (~70 net lines). No changes to
the controller, use-cases, types, SDK, or any backend/OAuth-provider config.


luanvdw (commented):
Thanks @sneub, this adds a solid QoL improvement and gives users a practical escape hatch when the CLI is running on one machine and the browser is on another machine! 

Could you/your agent review these small changes before we merge this in?

1. **Only enable the paste fallback when we can actually prompt**
If stdin is not a TTY, there is nowhere for the user to paste the callback URL. In that case, let’s keep the existing browser-callback behaviour, i.e. don’t start/race the paste flow, and don’t treat paste as an available fallback if opening the browser fails.

Put differently, the paste path should only change behaviour in interactive terminals.

2. **Make completion run once**
Since the browser callback _and_ pasted callback can both arrive, let’s centralize completion through a small `completeOnce(callbackUrl)` helper. That helper should make sure `state.handleCallback(...)` runs only once, and both paths await the same completion if they arrive close together.

3. **Make the remote-machine instruction more concrete**
I think “resulting callback URL” may be unclear to users. Could we make it more explicit
Suggested copy:

```text
Open this URL to sign in: <auth-url>

If the browser opens on another machine, finish sign-in there. When it redirects to localhost, copy the full localhost URL from the address bar and paste it here.
```

4. **Keep login alive after a bad paste**
If the user presses Enter too early or pastes the wrong thing, we should not end the whole login. Can we show a short message and ask again, while still leaving the browser callback path open? If this opens a big conditional loop, skip this for now - we can always return to this if needed.

sneub:
Thanks @luanvdw — all four addressed in 20dbbcb (single file `login.ts` + tests):

1. **Paste fallback gated on a TTY.** The paste prompt is now only created and raced when `stdin.isTTY` — non-TTY (CI, pipes, tests) keeps the existing browser-callback-only behaviour. Relatedly, the `openUrl` failure is only swallowed on a TTY; without one there's no paste escape hatch, so a failed browser launch now surfaces as an error again instead of hanging on a callback that'll never arrive.

2. **Completion runs once.** Both paths now funnel through a `completeOnce(callbackUrl)` helper backed by a shared in-flight promise, so `state.handleCallback(...)` runs at most once even when the browser redirect and the paste land together — and both paths await the same result. The promise is cleared on failure so a retry can still try again.

3. **Concrete instruction copy.** Adopted your suggested wording:
   > Open this URL to sign in: <auth-url>
   >
   > If the browser opens on another machine, finish sign-in there. When it redirects to localhost, copy the full localhost URL from the address bar and paste it here.

4. **Login stays alive after a bad paste.** A premature Enter, an unparseable paste, or a callback the exchange rejects now prints a short hint and re-prompts, with the browser callback path left open the whole time. It stayed contained (a small loop in one function), so I kept it in rather than deferring.

One CodeRabbit point I intentionally **didn't** act on: it flagged that the HTTP listener is torn down right after the race resolves, so a *late* browser redirect that arrives after the paste path already won won't get the success page. Once the paste path completes, the user is already signed in at their terminal — the only cost is a stray browser tab showing a connection error, which I judged not worth the deferred-shutdown complexity. Happy to revisit if you'd rather close that gap.

Tests added for: paste-only completion, run-once when both paths fire, re-prompt on bad paste / rejected exchange, and non-TTY browser-failure surfacing. Full suite green (232 tests).

luanvdw (approved):
Re-reviewed after branch update. Previous feedback is addressed, and checks are green.

PR #33, opened 2026-05-25
feat(cli): add app domain commands

AmanVarshney01:
## Summary
- add `prisma-cli app domain` commands for add, show, remove, retry, and wait
- wire the commands to the Management API custom-domain endpoints
- document the domain command/error contract and cover the key flows in tests

### DIFF
--- a/docs/product/command-principles.md
+++ b/docs/product/command-principles.md
@@ -51,6 +51,7 @@ The CLI should keep the meaning of these nouns stable:
 - `database`
 - `app`
 - `deployment`
+- `domain`
 
 If a noun means one thing in docs and a different thing in commands or output, the model is already drifting.
 
@@ -97,6 +98,14 @@ Build and release an app into a target branch.
 ### `logs`
 
 Resolve a deployment and show or stream its logs.
+
+### `wait`
+
+Block until a remote resource reaches a terminal state.
+
+Example:
+
+- `app domain wait`
 
 ### `promote`
 
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -771,6 +771,7 @@ Behavior:
 - is idempotent for a hostname already attached to the same app
 - does not re-trigger DNS verification for an existing row
 - prints DNS record instructions when available from the API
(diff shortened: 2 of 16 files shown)


luanvdw (commented):
Thanks @AmanVarshney01, the CLI shape is looking good!

Can we do an additional pass on DNS guidance? where possible, the CLI must always give a truthful next step after app domain `add/show`. IIUC right now it can omit records when the API omits `dnsRecords`, and it can print `CNAME` fixes for non-CNAME/non-DNS failures. Can we either rely on authoritative dnsRecords from the API, or make the fallback a tested provider contract?

I’d also fix the hardcoded "show" error label in shared hostname lookup. After that I think this achieves the [J20](https://linear.app/prisma-company/initiative/compute-public-beta-3290bbba60c0/overview) CLI objective without expanding scope. wdyt?

luanvdw (dismissed):
@AmanVarshney01 thanks for the follow-up changes, this is in a great shape!

Would you take a look at the below feedback, and see if we can either include it in this PR, or tackle as follow up work? I'm approving the PR so long, as these are non-blocking changes.

- Can we handle the `domain already registered elsewhere` as a more guided error path?
`addDomain` currently treats `409` as idempotent if the hostname is found on the selected app, which is good. But if the hostname is registered on a different app/project/workspace, it falls through to the generic “custom domain add failed” path and suggests `app domain show <hostname>`. Since `show` resolves within the selected app, that can send the user/agent into a loop: show says not found, add says conflict, etc. I’d make this a specific recovery message: “This hostname is already registered outside the selected app. Select the owning app, remove it there, or contact support.” We don’t need list/transfer/reassign in this PR, just a better dead-end.

2 smaller nits:

- `renderAppDomainShow` tones the failure row as error only when `failureReason` exists, but `formatDomainFailure` can show a `failureCategory` without a reason. If `failureCategory || failureReason` is present, the row should render as an error.
- can we make `retry` output a bit more informative? `app domain retry` currently prints hostname/status only. Since retry is part of the recovery path, it would be useful to include DNS records and/or the same failure/fix fields when present, similar to `show`. Not blocking, but it would make the command feel less like “now run show to understand what happened.”

luanvdw (dismissed):
Lessgo! 🚀

luanvdw (approved):

PR #34, opened 2026-05-25
fix(cli): place Next.js public/static next to nested server.js for mo…

sneub:
This PR pairs with: https://github.com/prisma/project-compute/pull/78

Deploying a Next.js app from inside a monorepo failed because the artifact's declared entrypoint and the on-disk location of `server.js` disagreed.

When `output: "standalone"` runs in a monorepo, Next preserves the workspace-relative path from the workspace root down to the app. So for an app at `apps/web`, `server.js` lands at `.next/standalone/apps/web/server.js`, not at the standalone root. The deploy pipeline assumed `server.js` was always at the root: the SDK reported `entrypoint: "server.js"` and the CLI copied `public/` and `.next/static/` to the artifact root, so the runtime couldn't find the server and static assets 404'd.

Fix:
- (in `@prisma/compute-sdk`, see [link to SDK PR]) the Next adapter now scans the standalone output for `server.js` (skipping `node_modules` to avoid third-party decoys) and reports a posix-joined entrypoint such as `apps/web/server.js`.
- This PR makes `restageNextjsArtifact` accept the `BuildArtifact`, derive the server subpath from `entrypoint`, and place `public/` and `.next/static/` next to `server.js` instead of at the artifact root.
- Add a regression test covering the monorepo case (server.js nested at `apps/web/server.js`, assets land under `apps/web/public/` and `apps/web/.next/static/`).

Single-app deploys (entrypoint `"server.js"`, derived subpath `""`) are unchanged.

### Coupling with the SDK
This change requires `@prisma/compute-sdk` ≥ 0.19 to take effect — that's the version that actually emits the nested entrypoint. Against 0.18.x the entrypoint is always `"server.js"`, the derived subpath is empty, and behavior matches the prior code. The dep bump can land in this PR or a follow-up.

### DIFF
--- a/packages/cli/src/lib/app/preview-build.ts
+++ b/packages/cli/src/lib/app/preview-build.ts
@@ -220,6 +220,7 @@ export async function restageNextjsArtifact(artifact: BuildArtifact, appPath: st
 }
 
 function nextjsServerSubpath(entrypoint: string): string {
+  // SDK emits posix-style entrypoints (path.posix.join).
   const dir = path.posix.dirname(entrypoint);
   return dir === "." ? "" : dir;
 }
--- a/packages/cli/src/lib/app/preview-build.ts
+++ b/packages/cli/src/lib/app/preview-build.ts
@@ -77,7 +77,7 @@ export async function executePreviewBuild(options: {
 
   try {
     if (buildType === "nextjs") {
-      await restageNextjsArtifact(artifact.directory, options.appPath);
+      await restageNextjsArtifact(artifact, options.appPath);
     }
 
     await normalizeArtifactSymlinks(artifact.directory, options.appPath);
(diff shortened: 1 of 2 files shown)


packages/cli/src/lib/app/preview-build.ts
```diff
@@ -192,23 +193,37 @@ async function restageNextjsArtifact(artifactDir: string, appPath: string): Prom
...
-    await cp(staticDir, path.join(artifactDir, ".next", "static"), {
+    await cp(staticDir, path.join(serverDir, ".next", "static"), {
       recursive: true,
       verbatimSymlinks: true,
     });
   }
 }
 
+function nextjsServerSubpath(entrypoint: string): string {
+  const dir = path.posix.dirname(entrypoint);
```
pcan, line 224: `path.posix.dirname` is correct here as long as the SDK always emits posix-style entrypoints — and it does today. Worth a brief comment or an explicit normalization (e.g. `entrypoint.replaceAll("\\", "/")`) to make the assumption visible. If the SDK's joining logic ever changes on Windows CI, this would silently produce a wrong `serverDir` rather than throwing.

sneub: Went with the comment option in 108d258 — added a one-liner inside `nextjsServerSubpath` documenting that the SDK emits posix-style entrypoints.

Skipped the `replaceAll("\\", "/")` normalization on purpose: since the SDK is same-org and already uses `path.posix.join` explicitly, a silent normalization would paper over a future contract drift rather than surface it (which I read as the opposite of your "rather than throwing" framing). Happy to escalate to an explicit `throw` if a backslash ever shows up, if you'd prefer the louder version.

pcan (approved):

PR #40, opened 2026-05-27
Use /v1/me for CLI auth identity

luanvdw:
## Summary

Updates `auth whoami` to resolve the signed-in identity through the Management API `/v1/me` endpoint.

- bumps `@prisma/management-api-sdk` to `^1.34.0`, which includes `/v1/me`
- calls `/v1/me` for stored OAuth sessions and `PRISMA_SERVICE_TOKEN` sessions
- prints OAuth users as `user: email@example.com`
- prints service-token sessions as `user: <service token: name>` or `user: <service token>`
- returns structured `user`, `workspace`, and `credential` in JSON output
- preserves the existing token-claim/workspace lookup fallback for rollout safety

## Validation

- `pnpm --filter @prisma/cli exec vitest run tests/auth-ops.test.ts tests/auth-real-mode.test.ts tests/auth-usecases.test.ts tests/auth.test.ts tests/project-usecases.test.ts`
- `pnpm --filter @prisma/cli test`
- `git diff --check`

## Notes

The Management API PR has landed and `@prisma/management-api-sdk@1.34.0` has been published, so this PR now consumes the generated `/v1/me` SDK types directly.


luanvdw:
Ran the updated `pr-review` flow against this CLI PR.

Verdict: conditional pass.

Findings:
- R1 — Auth result contract: `packages/cli/src/types/auth.ts` made `credential` optional even though the documented JSON contract says auth results include `credential` as an object or `null`. Optionality would let future code paths accidentally omit the field and drift the structured output contract.
- R2 — SDK type boundary / merge gate: `packages/cli/src/lib/auth/auth-ops.ts` currently uses a local `CurrentPrincipalClient` shim plus cast for `/v1/me`. That is acceptable while this PR is draft and the Management API SDK has not been regenerated, but before merge this should be replaced by the generated SDK type/version after the API PR lands.
- M1 — Presenter duplication: `packages/cli/src/presenters/auth.ts` duplicated the user-vs-credential label branch across login and whoami output, which creates a small drift point for future auth identity formatting.

File-size gate:
- No changed implementation or test file crosses the 1,000-line threshold.
- `docs/product/command-spec.md` is already over 1,000 lines; this PR adds a focused auth-contract section, not a new large code owner.

Validation limit:
- The repo typecheck currently fails on unrelated pre-existing errors in project presenter/helpers/publish-prep files, so I used tests plus diff hygiene for this pass.

luanvdw:
Follow-up from `pr-review`: code fixes pushed in `305db4e`.

- R1 fixed: made `AuthStateResult.credential` required as `AuthCredential | null`, and updated affected fixtures to keep the null case explicit.
- R2 accepted/no change for now: the `/v1/me` SDK shim remains because the Management API SDK release is still upstream of this draft PR. This should be removed when the generated SDK version is available before CLI merge.
- M1 fixed: centralized auth display labeling so OAuth user email and service/management token labels are formatted through one presenter helper.

Validation run:
- `pnpm --filter @prisma/cli exec vitest run tests/auth-ops.test.ts tests/auth-real-mode.test.ts tests/auth-usecases.test.ts tests/auth.test.ts tests/project-usecases.test.ts`
- `pnpm --filter @prisma/cli test` passed with local callback-server binding allowed
- `git diff --check origin/main...HEAD`

Validation note:
- `pnpm --filter @prisma/cli exec tsc --noEmit --pretty false` still fails on unrelated existing errors in `src/presenters/project.ts`, `tests/helpers.ts`, and `tests/publish-prep.test.ts`.

PR #41, opened 2026-05-27
fix(cli): drop dangling pnpm hoist symlinks from Next.js standalone build

sneub:
## Summary

Fixes Next.js standalone deploys that fail on pnpm monorepos because of dangling symlinks in `.pnpm/node_modules/`.

Next's standalone output reproduces pnpm's flat hoist layer (`node_modules/.pnpm/node_modules/<pkg>`), which routinely contains symlinks whose targets are absent — Next's require-trace doesn't always align with what pnpm actually populated in the store. The CLI's standalone-staging pass (`copyPathMaterializingSymlinks` → `resolveSymlinkTarget`) treated any dangling link as fatal and threw `Next.js standalone symlink target is missing: …`, failing the build.

After this change, dangling symlinks under `.pnpm/node_modules/` are silently skipped — the artifact simply has no entry at that path. Dangling symlinks **anywhere else** in the standalone tree still throw, so a genuinely missing top-level dependency isn't masked.

## What changed

- `resolveSymlinkTarget` now returns `string | null`. It returns `null` instead of throwing when a dangling link sits under the pnpm hoist layer.
- `copyPathMaterializingSymlinks` returns early when `resolveSymlinkTarget` returns `null`, so the broken link is never propagated to the artifact.
- New `isPnpmHoistLink` helper matches both flat (`.pnpm/node_modules/<pkg>`) and scoped (`.pnpm/node_modules/@scope/<pkg>`) paths via a path-component scan.
- The standalone source tree (`.next/standalone`) is untouched — we just don't propagate the broken link forward into the artifact that gets deployed.

## Why narrow, not broad

The original throw was a deliberate tripwire — better to fail loudly at build than ship a subtly-broken artifact. We're preserving that for top-level `node_modules/<pkg>` symlinks (the load-bearing ones Next emits from its require-trace). We're only relaxing it for `.pnpm/node_modules/<pkg>` — pnpm's flat hoist layer is structurally noisy and dangling entries there don't indicate anything wrong with the app.

### DIFF
--- a/packages/cli/src/lib/app/preview-build.ts
+++ b/packages/cli/src/lib/app/preview-build.ts
@@ -343,6 +343,9 @@ async function copyPathMaterializingSymlinks(
 
   if (sourceStat.isSymbolicLink()) {
     const resolvedTarget = await resolveSymlinkTarget(sourcePath, options);
+    if (resolvedTarget === null) {
+      return;
+    }
     await copyPathMaterializingSymlinks(resolvedTarget, destinationPath, options);
     return;
   }
@@ -376,7 +379,7 @@ async function resolveSymlinkTarget(
     appRoot: string;
     sourceRoot: string;
   },
-): Promise<string> {
+): Promise<string | null> {
   const linkTarget = await readlink(symlinkPath);
   const resolvedTarget = path.resolve(path.dirname(symlinkPath), linkTarget);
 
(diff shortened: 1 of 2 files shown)


luanvdw (approved):
Thanks @sneub!

One thing to consider, `isPnpmHoistLink` currently matches _any_ symlink path anywhere under a `.pnpm/node_modules` segment. The intent sounds narrower: skip only direct pnpm hoist-layer links (`.pnpm/node_modules/<pkg>` and `.pnpm/node_modules/@scope/<pkg>`) not sure if that matches what you had intended for this to do, so just flagging it for awareness :)

PR #48, opened 2026-05-28
fix: support current app deploy frameworks

AmanVarshney01:
## Summary
- detect current TanStack Start packages and `next.config.mts`
- support vanilla Bun deploys through `--framework bun` or `--entry`
- make official Hono scaffolds deploy without needing `--entry` by defaulting to `src/index.ts`
- remove the hidden deploy `--build-type` path so deploy uses `--framework`

## Checked
- `pnpm --filter @prisma/cli test`
- `pnpm --filter @prisma/cli build`
- official `create-next-app` deploy returned HTTP 200 with `Hello world!`
- official Hono, TanStack Start, and Bun local Compute builds passed

### DIFF
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -627,7 +627,7 @@ Behavior:
 - accepts repeated `--env NAME=VALUE` flags
 - maps user-facing framework names to deploy build strategies
 - uses `src/index.ts` as the Hono deploy entrypoint when the app has no `package.json#main` or `package.json#module` and that file exists
-- supports vanilla Bun apps with `--framework bun --entry <path>`
+- supports vanilla Bun apps with `--framework bun` using `package.json#main` or `package.json#module`, or with `--entry <path>`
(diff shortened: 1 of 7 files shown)


luanvdw (dismissed):
@AmanVarshney01 thanks for pushing this through. I pulled the branch locally and the implementation checks out: `pnpm --filter @prisma/cli test` and `pnpm --filter @prisma/cli build` both pass for me.

Two small things before we merge:
- Could you rebase against current main to resolve the current conflicts (let me know if I can help you with any of these) I think these conflicts are from the changes I made to explicit project setup. So I'm also happy to resolve these, if helpful. Just let me know
- small nit: `examples/next-smoke/README.md` still documents app deploy --build-type nextjs, but deploy now rejects --build-type. Can you switch that example and the related bullet to `--framework nextjs`?

Approving so long, so you're unblocked from merging afterwards

luanvdw (approved):

PR #62, opened 2026-06-02
Document CLI I/O timeout behavior

rtbenfield:
Draft PR for visibility only. This contains the CLI I/O timeout spec and implementation plan; it is not intended to be merged.

@luanvdw tagging you for awareness.

## Changes

- **Timeout spec**: Adds `.agents/projects/cli-io-timeouts.spec.md`, defining the intended CLI behavior for stalled Prisma-controlled I/O, including `OPERATION_TIMEOUT`, cancellation separation, generous defaults, and out-of-scope boundaries.
- **Implementation plan**: Adds `.agents/projects/cli-io-timeouts.plan.md`, sequencing the work through a shared timeout contract, Prisma API/SDK request boundaries, long-lived inactivity boundaries, and product-doc alignment.
- **Command inventory**: Adds pseudocode callstacks for each current command showing where request, wait, stream-inactivity, existing SDK, or no timeout should sit in the execution hierarchy.

## Why

The spec chooses step-scoped and inactivity-scoped timeouts instead of a root command-wide timeout. That keeps long-running workflows such as `app logs`, `app domain wait`, `auth login`, deploys, and local app runs usable while still preventing Prisma-controlled API, SDK, polling, and stream boundaries from hanging forever.

### DIFF
--- a/.agents/diary/cli-io-timeouts.md
+++ b/.agents/diary/cli-io-timeouts.md
@@ -1,0 +1,5 @@
+[Out-of-scope decision] Excluded a global public `--timeout` flag from the first slice because request, inactivity, and total wait deadlines mean different things across commands. The spec keeps public configuration narrow until usage shows which budgets users need to tune.
+
+[Out-of-scope decision] Excluded automatic retry policy changes because bounded waiting and retry behavior should be designed separately. Mixing them in this spec would obscure the core product contract: stalled I/O must fail clearly and predictably.
+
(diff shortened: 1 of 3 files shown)


wmadden-electric:
Closing as stale: this documents the I/O timeout behaviour of the commander shell, which was deleted in #139 — the behaviours it describes (controller-level fetch timeouts, the old streaming paths) no longer exist as written. The engine has its own timeout story (for example the bounded identity enrichment in `auth whoami`), and documenting it belongs to the engine docs when the streaming transport lands. The research in the .agents notes remains readable from this PR's branch if anyone picks that up.

PR #65, opened 2026-06-03
fix: resolve env list from active branch

luanvdw:
## Overview

This changes `project env list` so an unscoped read reflects the caller's current workstream instead of silently showing production. It keeps writes explicit, but makes read output clear about which env map was resolved and why.

## Changes

- `packages/cli/src/lib/git/local-branch.ts` extracts local Git branch detection from app deploy and is reused by `packages/cli/src/controllers/app.ts` and `packages/cli/src/controllers/app-env.ts`.
- `runEnvList` now has a list-only target resolver. Explicit `--role` and `--branch` paths stay unchanged; no flags with a local Git branch resolve the matching Platform Branch role; missing Platform Branches show preview template metadata; no Git branch shows a production/preview project-level overview.
- `packages/cli/src/types/app-env.ts` and `packages/cli/src/presenters/app-env.ts` add resolved target metadata so human and JSON output can explain the inferred branch, branch role, env map, and missing-branch state.
- Env-var listing now uses one shared collector for pagination and filtering. The overview path excludes Branch overrides by keeping only project-level production and preview rows.
- Docs, help examples, and env-list tests were updated to cover preview, production, missing Branch, no-Git overview, and explicit-scope behavior.

## Why

A default production read is too easy to misinterpret when the user is checked out on a feature branch. Resolving from local Git matches the Branch model users already carry in their head, while the no-Git overview avoids pretending there is a single safe target. The implementation uses Branch `role` as the platform truth and does not couple it to `isDefault`; production-branch mutation remains out of scope.

### DIFF
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -687,7 +687,9 @@ Every write targets exactly one scope:
 - `--role` and `--branch` are mutually exclusive.
 - For write verbs (`add`, `update`, `remove`), one scope flag is required
   so the CLI never silently writes to production.
-- For read verbs (`list`), omitting `--role` defaults to `--role production`.
+- For read verbs (`list`), omitting `--role` or `--branch` resolves from
+  the active local Git branch when one exists; outside a Git branch it
+  shows a production/preview project-level overview.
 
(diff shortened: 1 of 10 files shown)


luanvdw:
@coderabbitai review

PR #74, opened 2026-06-08
Support first deploy database setup

luanvdw:
## Overview

Extends app deploy --db from preview-only branch databases to the first production deploy path, while keeping production guardrails around existing env vars and live deployments.

## Changes

- Generalizes deploy DB setup so preview writes branch env overrides and first production setup writes production DATABASE_URL / DIRECT_URL through the same env-var APIs.
- Uses the production deploy gate to identify whether the selected production App has no live deployment before allowing setup.
- Treats existing production DATABASE_URL or DIRECT_URL as BYO DB intent and skips without overwriting.
- Updates CLI/help/product docs and adds controller tests for production bootstrap guardrails.

## Why

This keeps the preview branch model and avoids managed bindings. Production setup only runs before the App has a live deployment, and any existing DB env var is treated as explicit BYO DB intent.

## Validation

- pnpm --filter @prisma/cli test tests/app-branch-database.test.ts
- pnpm --filter @prisma/cli test tests/production-deploy-gate.test.ts
- pnpm --filter @prisma/cli test
- pnpm build:cli
- git diff --check


AmanVarshney01 (approved):
I checked the PR with create-prisma flow and it works great

PR #79, opened 2026-06-09
Add preview environment variable pull command

luanvdw:
## Overview
Adds a CLI-only first slice for pulling preview environment variables into a local dotenv file. This lets local development use platform preview values without printing secret values to terminal or JSON output.

## Changes
- Adds `project env pull [output-file]` with `--role preview`, `--branch`, `--project`, default `.env.local`, and command/help metadata.
- Wires the command to the future `POST /v1/environment-variables/pull` endpoint and returns `FEATURE_UNAVAILABLE` when that endpoint is not available.
- Adds file safety in `packages/cli/src/lib/app/env-pull-file.ts`: rejects tracked files, outside paths, non-file targets, and missing parents; requires confirmation or `--yes`; writes dotenv output with `0600` permissions.
- Updates product docs/readmes and tests role, branch, overwrite, tracked-file, missing-endpoint, and no-value-leak paths.

## Why
Preview values are local dev/test configuration, so the CLI can pull them only through an explicit file-writing command. Production values remain write-only. Keeping the API call behind a small wrapper lets this CLI-only PR land before backend support is available.

## Validation
- `pnpm --filter @prisma/cli exec vitest run tests/app-env.test.ts tests/app.test.ts --reporter=verbose`
- `pnpm build:cli`
- `pnpm --filter @prisma/cli test` outside the sandbox because auth tests bind `127.0.0.1`
- `git diff --check`
- Local review skills: `tmp-code-review`, `thermonuclear-review`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ pnpm prisma-cli auth login
 pnpm prisma-cli app deploy --env DATABASE_URL=postgresql://example
 pnpm prisma-cli project env add --file .env --role preview
 pnpm prisma-cli project env list --role preview
+pnpm prisma-cli project env pull
 ```
 
 If you want local project scripts that look like the future command shape, add:
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -862,8 +862,9 @@ prisma-cli app deploy --entry src/server.ts --http-port 3000
 Manage durable, platform-stored environment variables for the resolved
 project. The `env` namespace operates on the
 platform-managed `/v1/environment-variables` API; values are stored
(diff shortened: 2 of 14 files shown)


wmadden-electric:
Closing as stale: `env pull` was built on the commander shell's env controllers and command registration, all deleted in #139. The engine CLI ships `project env add|update|list|remove`; pulling variables into a local dotenv file did not make the port and is now recorded in `.drive/projects/prisma-cli-v8/deferred.md`. If wanted, it returns as an engine command — note the current `project env list` deliberately never returns values, so an engine `env pull` needs its own ruling on secret handling before it is built.

PR #80, opened 2026-06-09
Adopt typed result errors for project resolution

rtbenfield:
Converts project target resolution to return typed better-result errors instead of throwing expected `CliError` instances, while preserving the existing command-facing error envelopes.

## Changes

- **Project resolution**: `packages/cli/src/lib/project/resolution.ts` now models project not found, ambiguous project, setup required, stale local state, and workspace mismatch as `TaggedError` variants returned through `Result` composition.
- **Command boundaries**: Project, app, branch, database, and app env controllers now convert resolution errors exhaustively back into the existing `CliError` envelopes at the command boundary.
- **Tests and plan tracking**: Resolver tests assert typed result failures plus stable public conversion, and phase 3 is marked complete in the adoption plan with the remaining unrelated typecheck blocker documented.

## Why

This keeps expected project-resolution failures typed below the CLI boundary and makes new variants require exhaustive conversion before they can reach command output. Setup validation remains on the existing compatibility path for the next phase, avoiding a broader migration than phase 3 requires.

## Verification

- `pnpm --filter @prisma/cli exec vitest run tests/project-resolution.test.ts tests/project.test.ts tests/app-controller.test.ts tests/branch.test.ts tests/branch-controller.test.ts`
- `pnpm --filter @prisma/cli exec tsc -p tsconfig.json` currently fails on unrelated existing database/test typing errors documented in `.agents/diary/better-result-phase-3.md`.

### DIFF
--- a/.agents/diary/better-result-phase-3.md
+++ b/.agents/diary/better-result-phase-3.md
@@ -1,1 +1,3 @@
-+ Scope tangent worth pursuing later: Package typecheck now runs after installing dependencies, but it reports unrelated existing errors in `branch-database.ts`, test helper typings, project real-mode mock types, and script module declarations. I did not fix these because phase 3 is limited to project resolution domain errors.
+# Better Result Phase 3
+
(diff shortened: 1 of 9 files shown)


luanvdw (approved):
🙏

PR #81, opened 2026-06-10
Add compute runtime sleep guard package

rtbenfield:
Adds `@prisma/compute` as a publishable runtime utility package for Prisma Compute applications. The first API helps application code keep an instance awake while background work is active, without adding any CLI command surface.

## Changes

- **Compute package**: Adds `packages/compute` as an ESM TypeScript library with `tsdown` build output, package-local verification guidance, and workspace lockfile metadata.
- **Sleep guard utilities**: Exposes `ScaleToZeroGuard` and `waitUntil` from `packages/compute/src/index.ts`. Both write paired acquire/release signals to the Prisma Compute runtime endpoint, no-op outside the runtime, and support `AbortSignal` as a safety bound for dangling guards.
- **Consumer docs and tests**: Adds README guidance for human authors and coding agents, plus tests covering signal writes, idempotent release, disposable cleanup, abort-triggered release, unavailable runtime no-op behavior, and `waitUntil` fire-and-forget semantics.

## Why

Prisma Compute can sleep and resume applications from memory snapshots after short idle periods. Request-adjacent background work needs an explicit keep-awake signal so it is not interrupted when the request completes. The guard uses immediate one-byte writes so acquisition and release are ordered with user code, while `AbortSignal.timeout(ms)` lets callers bound guard lifetime if a code path does not reach release.

### DIFF
--- a/.github/workflows/preview-cli-package.yml
+++ b/.github/workflows/preview-cli-package.yml
@@ -58,9 +58,17 @@ jobs:
         id: cli_tests
         run: pnpm --filter @prisma/cli test
 
+      - name: Run compute tests
+        id: compute_tests
+        run: pnpm --filter @prisma/compute test
+
       - name: Build CLI package
         id: cli_build
         run: pnpm --filter @prisma/cli build
+
+      - name: Build compute package
+        id: compute_build
+        run: pnpm --filter @prisma/compute build
 
       - name: Prepare staged preview package
         id: prepare_preview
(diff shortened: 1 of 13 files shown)


luanvdw:
Reviewed with the local code-review. Fixed the waitUntil rejection handling, added a regression test, and made the PR workflow run compute test/build too. Local checks and CI are green.

PR #87, opened 2026-06-12
Fix npm trusted publishing workflows

rtbenfield:
Restores the publish workflows to the npm CLI path required for npm trusted publishing while keeping pnpm for repo setup and package preparation.

## Changes

- **Publish workflows**: Use Node 24 and configure the npm registry through setup-node for the CLI and Compute publish jobs.
- **Release publishing**: Switch final package publication back to npm publish so npm can perform the trusted-publisher OIDC authentication flow.
- **Documentation**: Record the npm publish exception in AGENTS.md and update the publishing ADR command example.

## Why

npm trusted publishing is implemented by the npm CLI OIDC exchange. Keeping pnpm for install, versioning, and pack validation is fine, but the final publish step needs npm publish to avoid falling back to unauthenticated registry publishing.

### DIFF
--- a/.github/workflows/publish-cli.yml
+++ b/.github/workflows/publish-cli.yml
@@ -36,7 +36,8 @@ jobs:
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version-file: .node-version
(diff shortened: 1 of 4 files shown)


rtbenfield:
@coderabbitai full review

PR #89, opened 2026-06-12
Rename compute keep-awake guard

rtbenfield:
Clarifies the keep-awake safety behavior in @prisma/compute and renames the scoped guard API away from scale-to-zero terminology.

## Changes

- **Public API**: Renames `ScaleToZeroGuard` and `ScaleToZeroGuardOptions` to `KeepAwakeGuard` and `KeepAwakeGuardOptions`, including the source and test filenames.
- **Documentation**: Updates README and JSDoc wording to make the `AbortSignal` behavior explicit: aborting releases the keep-awake guard only and does not cancel the underlying promise or guarded work.
- **Tests**: Updates compute package tests to use the new public guard name while preserving existing behavior coverage.

## Why

`KeepAwakeGuard` describes the caller-visible behavior more directly than scale-to-zero terminology. The signal wording is intentionally explicit because the abort signal is a cost safety net for releasing the guard, not a cancellation mechanism for the work being waited on.


rtbenfield:
@coderabbitai full review

PR #91, opened 2026-06-18
fix(app): resolve management-command branch to the project default

kristof-siket:
## Problem

`app show`, `logs`, `list-deploys`, `open`, `promote`, `rollback`, and `remove` scope the app lookup to a single branch via `GET /v1/compute-services?projectId=…&branchGitName=…`. The branch is resolved as `--branch` → active git branch → literal `main` (per `command-spec.md`).

The management API returns an empty page (HTTP 200, `data: []`) when `branchGitName` matches no live branch. So an app deployed via git-push whose default/production branch is not `main` — e.g. `master` — is invisible to these commands unless the caller happens to be on a local `master` checkout:

```
$ prisma-cli app logs --project proj_…
✘ No deployments available to stream logs [NO_DEPLOYMENTS]
```

The control plane and management API are correct; the app, its branch (`master`, `isDefault: true`), and its versions all exist. The defect is the CLI's branch resolution.

## Fix

Management commands never create branches, so they now resolve the branch they read against the project's actual branches:

1. `--branch <name>` — honored as-is
2. the active git branch — only when a branch with that name exists in the project
3. the project's default (production) branch — from `GET /v1/projects/:id/branches`

This drops the implicit `main` fallback for read commands. Deploy is unchanged.

- `src/lib/app/read-branch.ts` (new): read-only `resolveReadBranch` — match by name, else default branch, else null. Lives in its own module so the existing `preview-provider` test mocks don't need to stub it.
- `src/controllers/app.ts`: `resolveProjectContext` resolves the inferred branch via the API.
- `docs/product/command-spec.md`: branch-resolution spec updated (the `main` fallback was documented; management commands now resolve existing-or-default).

## Testing

- `tests/read-branch.test.ts` (new): match / default-fallback / no-branches.
- Updated the shared `verboseContext.branch` expectation in `tests/app-controller.test.ts` (now carries the resolved branch id + role).
- Full suite: **484 passing**. `tsc --noEmit` clean. Biome clean on changed lines.

## Note

Even once merged, `bunx @prisma/cli@latest` only picks this up after a release.


kristof-siket:
## ✅ Live verification report

Exercised the affected `app` commands end-to-end against the **real production project** that originally surfaced the bug, using my existing CLI login (`api.prisma.io`). All commands are **read-only** (`show`, `list-deploys`, `logs`) — no mutations.

- **Project:** `proj_cmqi1wcoe02ei1mf5c392aoyi` — git-push app `voice-vault-back-end`, default/production branch **`master`**.
- **Baseline:** `bunx @prisma/cli@latest` (`3.0.0-beta.11`, pre-fix).
- **Fixed:** this branch, run from source via `tsx packages/cli/src/bin.ts`.

### Before / after (same project, same auth)

| Command | Baseline `@latest` (beta.11) | Fixed (this branch) |
| --- | --- | --- |
| `app show` (no `--app`) | `app: null`, `recentDeployments: []` ❌ | finds ≥1 app (prompts to select) ✅ |
| `app show --app voice-vault-back-end` | *would* `USAGE_ERROR` (empty candidate set) | app + `liveUrl` + live deploy `cpv_cmqi2ccdi…` + both versions ✅ |
| `app list-deploys --app voice-vault-back-end` | `USAGE_ERROR: "Selected app does not exist in the resolved project"` ❌ | 2 deployments, `count: 2` ✅ |

Baseline `app show` output (the bug):

```json
{ "ok": true, "command": "app.show",
  "result": { "app": null, "liveDeployment": null, "liveUrl": null, "recentDeployments": [] } }
```

Fixed `app show --app voice-vault-back-end`:

```
app:            cps_cmqi1wncb00da0ddxcuzt65mu  (voice-vault-back-end)
liveUrl:        https://cmqi1wncb00da0ddxcuzt65mu.fra.prisma.build
liveDeployment: cpv_cmqi2ccdi065eygfacryq4fng
recent:         [cpv_cmqi2ccdi065eygfacryq4fng, cpv_cmqi1wowu0a90w0fa6iqqp59z]
```

### Both resolution paths exercised

- **Default-branch fallback** (ran from a `fix/…` checkout → inferred branch is **absent** in the project → falls back to the default branch `master`): app found ✅ — this is the fix.
- **Match path / no-regression** (ran from a `master` checkout → inferred branch **matches** an existing branch directly, the previously-working case): app found ✅ — behavior unchanged.

### Data fidelity

The fixed CLI returns exactly the ground-truth rows for the project (service `cps_cmqi1wncb…65mu`, versions `cmqi2ccdi…` + `cmqi1wowu…`, endpoint `…fra.prisma.build`) — it resolves the **correct** app, not merely a non-empty result.

`app logs` resolves through the same `resolveProjectContext → listApps → resolveExistingAppSelection → live-deployment` path that `app show` exercised here (it located live deployment `cpv_cmqi2ccdi…`), differing only in the final stream step.

### Conclusion

- **New behavior works:** a git-push app on a non-`main` default branch is now visible to the management commands regardless of the local Git branch.
- **No regression:** when the inferred branch exists in the project it resolves directly as before; only the previously-broken case (inferred branch absent) changed — from "empty" to "default branch". Consistent with the 484 passing unit/integration tests.

_Posted for the record (verification was run post-merge, read-only)._

PR #92, opened 2026-06-19
fix(app): resolve the custom-domain production branch by role, not name

AmanVarshney01:
## Problem

Custom-domain commands (`app domain add/show/remove/retry/wait`) attach to the **production branch**, but they identified it by the **literal name** `production`:

- `toBranchKind(name)` only returns `"production"` for a branch named `production` or `main`.
- The flow defaulted the branch to the literal string `"production"` and queried by that name.

So a project whose production branch is named anything else — e.g. **`master`** (a git-push deploy on `master`) — could never attach a custom domain. `domain add` looked for a branch named `production`, found none, and failed with a confusing *"no app on the production branch."* This is the same name-vs-role flaw as #91, on the domain path.

## Fix

Resolve the production branch from the project's branches **by role** (the durable default branch) and use its real name. The production-only gate now checks the **resolved branch's role**, not a guess from its name.

- **`read-branch.ts`** — extract `listProjectBranches`; add `resolveProductionBranch` (matches an explicit `--branch` by name so its role can be validated, else returns the production-role / default branch).
- **`resolveProjectContext`** — a `productionBranch` mode resolves the branch by role.
- **`resolveAppDomainTarget`** — drop the early name-based gate; gate on the **resolved** branch's role after resolution. Use the resolved branch's real name for the app lookup.

So `domain add` on a `master`-default project now finds `master` (role production) and attaches the domain there. An explicit `--branch <preview>` is still rejected — now by its **actual** role, not its name.

## Testing

- The branch mock now marks the default (durable) branch as `production` role (it was unrealistically `preview`).
- New: domain resolves a **`master`-named** production branch by role and attaches the domain.
- New: domain rejects a **real preview** branch by its role.
- Full suite **485 passing**, `tsc --noEmit` clean, biome clean.

### DIFF
--- a/packages/cli/src/controllers/app.ts
+++ b/packages/cli/src/controllers/app.ts
@@ -2118,11 +2118,19 @@ async function resolveAppDomainTarget(
 function resolveDomainBranch(
   explicitBranchName: string | undefined,
 ): ResolvedDeployBranch {
+  const branchName = explicitBranchName?.trim();
+  if (branchName) {
+    return {
+      name: branchName,
+      annotation: BRANCH_FLAG_ANNOTATION,
+      explicit: true,
+    };
+  }
+
   return {
-    name: explicitBranchName?.trim() || "production",
-    annotation: explicitBranchName
-      ? BRANCH_FLAG_ANNOTATION
-      : "production default",
+    name: "production",
+    annotation: "production default",
+    explicit: false,
   };
 }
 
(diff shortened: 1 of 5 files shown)


wmadden-electric:
Closing as stale, but the bug it fixes is real and survives. This PR made the old `app domain` code resolve the production branch by ROLE rather than by the literal name "production". The old controller is deleted (#139) — but the ported service code inherited the by-name check: `src/commands/service/target.ts` still derives branch kind from the name (`toBranchKind`, used at the same decision point this PR fixed). A project whose production branch is named e.g. `master` still cannot attach a custom domain. Recorded as a live bug in `.drive/projects/prisma-cli-v8/deferred.md`; the fix should land against `target.ts` with the branch role taken from the API record, as this PR did for the old code.

PR #93, opened 2026-06-19
feat(auth): support workspace session switching

AmanVarshney01:
## Summary
- add local multi-workspace OAuth session support with active workspace selection
- add auth workspace list/use/logout commands plus auth logout --workspace shortcut
- keep service-token precedence explicit while showing cached OAuth workspaces
- improve project mismatch guidance to switch workspaces instead of re-login

## Verification
- pnpm --filter @prisma/cli test -- auth.test.ts auth-real-mode.test.ts token-storage.test.ts
- pnpm --filter @prisma/cli exec tsc --noEmit
- pnpm --recursive exec tsc --noEmit
- pnpm lint
- git diff --check

### DIFF
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -68,11 +68,11 @@ The CLI accepts two authentication sources, in this fixed precedence:
 1. `PRISMA_SERVICE_TOKEN` environment variable — long-lived service token, intended for CI and other headless contexts.
 2. Stored OAuth session — created by `prisma-cli auth login`, kept in the OS-appropriate credentials store, refreshed automatically.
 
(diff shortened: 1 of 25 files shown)


luanvdw (commented):
This is great @AmanVarshney01 I’m excited for this!

One thing to consider,

`packages/cli/src/adapters/token-storage.ts:235`

`clearTokens()` still clears the credentials file and auth-context file outside `withRefreshLock()`. The PR does a good job serializing refresh writes plus `auth workspace use/logout`, but full `auth logout` is still a concurrent writer to the same files. That means a logout can race with an in-flight refresh or workspace mutation and leave credentials/context resurrected or partially inconsistent, which violates the new documented contract that `auth logout` clears all local OAuth workspace sessions.

Should we wrap full logout in the same lock, e.g. make `clearTokens()` call `withRefreshLock(() => clearTokensUnlocked())`, mirroring the pattern used for `setTokens`, `useWorkspace`, and `logoutWorkspace`.

luanvdw (approved):
🚀

PR #104, opened 2026-06-30
fix(build): predictable error handling + transient-read retry for build logs

kristof-siket:
## What

Two fixes to the `build logs <build_id>` command's failure behavior, stacked on top of #102 (now merged to `main`).

### 1. Predictable error handling — 401 maps to the shared auth error

The command streams `GET /v1/builds/{buildId}/logs` through openapi-fetch, which does **not** throw on non-2xx — a 401 arrives as `response.status === 401`. The command-runner's `SDKAuthError → authRequiredError` mapping therefore never fired for this path, so an expired/invalid CLI session surfaced as a generic `BUILD_LOGS_FAILED` ("Failed to read build logs… / Retry").

A 401 now maps to the shared `authRequiredError(["prisma-cli auth login"])`, so the output is consistent with the rest of the CLI and tells the user what to actually do. `404 → BUILD_NOT_FOUND` and the generic fallback are unchanged; 403 is intentionally not special-cased (the API uses 404-collapse for unauthorized access).

### 2. Bounded retry honoring the API's `retryable` signal, resuming from cursor

The read (stream open + NDJSON consumption) now runs inside a bounded retry loop:

- **3 attempts max**, backing off ~500ms then ~1500ms, abort-aware (a cancel during the wait returns cleanly, not as a failure).
- **Retries** on: a transient open status (`408, 429, 500, 502, 503, 504`), a non-Abort network-style error from the GET/stream read, or the Management API's existing terminal `error` record with `retryable: true`.
- **Surfaces immediately** (no retry): `401 → authRequiredError`, `404 → BUILD_NOT_FOUND`, any other non-transient status `→ BUILD_LOGS_FAILED`, or a terminal `error` with `retryable: false`.
- **Resumes from the last cursor** (tracked across log records and the terminal record) on every reconnect, so resumed reads don't reprint already-emitted output.
- On exhaustion after a retryable failure, behavior is unchanged from before: the terminal error message prints to stderr and `process.exitCode = 1` (terminal-record path), or the generic `CliError` is thrown (transient-open / network path).
- `--follow` keeps working: a retryable drop reconnects from cursor through the same bounded loop (the attempt count stays bounded — follow does not retry unbounded).

This fixes the intermittent **408-from-Durable-Streams** "Failed to read build logs." seen right after a build completes.

## Trace evidence

A real failure showed the two-failure pattern this PR targets:

1. First attempt: **401** — the CLI JWT had expired. (Now → actionable `authRequiredError` instead of a generic read failure.)
2. Second attempt: the route returned **200**, but the streams server (`GET …/v1/stream/build-logs`) returned **HTTP 408** after ~5.2s. (Now → a retryable transient that the loop reconnects from the last cursor.)
3. Third attempt: succeeded.

## Tests

New `packages/cli/tests/build-logs-controller.test.ts` (deterministic, backoff injected as zeros — no real timers):

- 401 → `authRequiredError` (asserts `AUTH_REQUIRED`, not `BUILD_LOGS_FAILED`).
- Retryable terminal error on attempt 1, success on attempt 2 → succeeds, prints both log batches, and the second read carries the resumed `cursor`.
- Retryable failure on all 3 attempts → `process.exitCode = 1`, message surfaced exactly once.
- Transient open status (503) then success → retried.
- `--follow` drop → reconnects from cursor with `follow=true` preserved.
- 404 and non-retryable terminal `end`/`no_logs` → **not** retried (unchanged).

`pnpm --filter @prisma/cli test` (562 tests) green; `tsc --noEmit`, `tsdown` build, and `biome check` all clean.

## Notes

- This was originally branched off `feat/build-logs-command` (#102). Since #102 merged to `main` first, the branch is rebased onto `main` and targets `main` — the diff is exactly these two files (no #102 diff). It can still be folded into the #102 line of work if preferred.
- No new generic retry framework: the loop, the abort-aware `sleep`, and the small phase helpers (`openBuildLogStream` / `consumeBuildLogStream`) are local to `build.ts`.

### DIFF
--- a/packages/cli/src/controllers/build.ts
+++ b/packages/cli/src/controllers/build.ts
@@ -31,10 +31,39 @@ export interface BuildLogsOptions {
   cursor?: string;
 }
 
+/**
+ * Backoff between retry attempts. The length also bounds the retry budget: with
+ * two delays the stream is opened at most three times. Injectable so tests can
+ * exercise the retry paths without real timers.
+ */
+const RETRY_BACKOFF_MS: readonly number[] = [500, 1500];
+
+const TRANSIENT_OPEN_STATUSES = new Set([408, 429, 500, 502, 503, 504]);
+
+type TerminalRecord = Extract<BuildLogRecord, { type: "terminal" }>;
+
+type BuildLogsClient = NonNullable<
+  Awaited<ReturnType<typeof requireComputeAuth>>
+>;
+
+interface BuildLogsDeps {
+  backoffMs?: readonly number[];
+}
+
+type ReadOutcome =
+  | { kind: "done" }
+  | { kind: "auth" }
+  | { kind: "not-found" }
+  | { kind: "fatal"; status: number }
+  | { kind: "retryable-open"; status: number }
+  | { kind: "retryable-network" }
+  | { kind: "retryable-terminal"; record: TerminalRecord };
+
 export async function runBuildLogs(
   context: CommandContext,
   buildId: string,
   options: BuildLogsOptions = {},
+  deps: BuildLogsDeps = {},
 ): Promise<void> {
   const client = await requireComputeAuth(
     context.runtime.env,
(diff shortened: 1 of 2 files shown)


kristof-siket:
@coderabbit review

wmadden-electric:
Closing as stale: this hardens `build logs` in `src/controllers/build.ts`, which was deleted when the Commander shell was retired (#139). The command was rewritten on the engine (`src/commands/build/logs.ts`) with structured error handling and its first-ever test suite, which covers the predictable-error half of this PR. The other half — retrying transient read failures mid-stream — was not carried over and is now recorded in `.drive/projects/prisma-cli-v8/deferred.md` alongside the engine's other streaming follow-ups (the stream also cannot yet settle non-zero on a failed build, which that file tracks too).

PR #105, opened 2026-07-01
feat(app): add --no-promote for promotionless deploys

kristof-siket:
## What

Adds a `--no-promote` flag to `prisma-cli app deploy` — a **promotionless deployment** that builds a new deployment **without** flipping the live pointer. Deploy stays promote-by-default; this is strictly opt-in.

When `--no-promote` is set:

- The new deployment is built, started, and polled to `running`, but **not promoted**.
- The **previous** deployment keeps serving live (its own URL and the App's live pointer are unchanged).
- The new deployment is "built, not live" and reachable only at its **own candidate URL**.
- A later `prisma-cli app promote <deployment-id>` makes it live (existing command, unchanged).

`promote` / `rollback` / `list-deploys` / `show-deploy` / `domain` and deploy's default behavior are untouched.

## Why — the CI build-then-verify path

CI wants to build a candidate, **health-check it** against a stable URL, and only then flip production. Promote-on-build gives CI no window to verify before customers see the new code. `--no-promote` splits the two steps:

```bash
prisma-cli app deploy --no-promote --json   # build candidate, get deployment.url
#  → curl / smoke-test result.deployment.url (the candidate's own URL)
prisma-cli app promote <deployment-id>       # flip production once green
```

Because a promotionless deploy **never replaces the live deployment**, it **bypasses the production-confirmation gate** (`enforceProductionDeployGate` → `PROD_DEPLOY_REQUIRES_FLAG`): `app deploy --no-promote` on the **production** branch builds a candidate **without `--prod`** and without a confirmation prompt.

## Grounding — SDK already supports it (no SDK / management-api change)

`@prisma/compute-sdk` `ComputeClient.deploy(options)` accepts `skipPromote?: boolean` (`DeployOptions` in `compute-client.d.ts`). With `skipPromote: true`, `deploy()` builds + starts + polls to `running` and returns:

```
{ ..., deploymentId, deploymentEndpointDomain, appEndpointDomain: null,
  promoted: false, previousDeploymentId, previousDeploymentAction: "still-active", ... }
```

`deploymentEndpointDomain` is the candidate's own URL (what CI health-checks); on the promoted path the SDK instead returns `appEndpointDomain: <live url>, promoted: true`. This PR only wires the CLI onto that existing SDK behavior.

## Changes

- **`lib/app/app-provider.ts`** — `deployApp` accepts `skipPromote`, forwards it to `sdk.deploy(...)`, and maps the return by `deployed.promoted`. When not promoted: `app.liveDeploymentId = previousDeploymentId` (the still-live old deployment), `deployment.live = false`, `promoted = false`. `DeployRecord` gains `promoted` and `deployment.live`. (The `liveUrl` / `deployment.url` expressions are unchanged from before — `appEndpointDomain` is null on the not-promoted path, so they already resolve to the candidate endpoint.)
- **`controllers/app.ts`** — threads `noPromote` into `deployApp({ skipPromote })`; skips `enforceProductionDeployGate` when `noPromote` (treating `firstProductionDeploy` as false); surfaces `promoted` + `deployment.live`; `nextSteps` lead with `app promote <id>` when not promoted.
- **`commands/app/index.ts`** — adds the `--no-promote` Commander option (same `--no-*` negatable idiom as the existing `--no-db`; exposes `options.promote === false`) and threads `noPromote`.
- **`presenters/app.ts`** — human output gains a "built, not live" branch: `Built <id> in <duration> (not promoted)`, the candidate URL, a note that the live deployment is unchanged, and a `Promote` hint. The promoted path is unchanged.
- **`types/app.ts`** — `AppDeployResult` gains `promoted` and `deployment.live`. `--json` (`serializeAppDeploy`) surfaces `result.promoted` and the candidate `result.deployment.{id,url,live}` (field name `url` kept, so existing consumers reading `.result.deployment.url` get the candidate URL under `--no-promote`).
- **`docs/product/command-spec.md`** — documents the flag (signature, behavior, gate bypass, output, deploy-all propagation, example), per this package's doc-driven-development rule.

## Tests

Extended the existing suites (562 passing, +7 new):

- **provider**: `deployApp({ skipPromote: true })` forwards it to `sdk.deploy` and maps the promotionless return (candidate URL, `deployment.live: false`, `promoted: false`, live pointer stays on `previousDeploymentId`); plus the promoted mapping.
- **controller**: `--no-promote` calls `deployApp` with `skipPromote: true`, returns `promoted: false` + candidate URL + `deployment.live: false`, and `nextSteps` include `app promote <id>`; and `--no-promote` on the production branch **does not** call `listDeployments` (gate bypassed) and deploys without `--prod`.
- **CLI layer**: `--no-promote` threads `noPromote: true` into `runAppDeploy` and `--json` surfaces `promoted: false` + the candidate `deployment.url`.
- **presenter**: not-promoted human output (built-not-live + promote hint) and JSON shape.

`pnpm --filter @prisma/cli test` (562 pass), `tsc --noEmit`, and `biome check` all pass; CLI builds; `node dist/cli.js app deploy --help` shows `--no-promote`.

> Note: the CI rewrite in `pdp-control-plane` that consumes this flag is a separate follow-up, gated on the CLI release.

### DIFF
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -1012,7 +1012,7 @@ prisma-cli app run --build-type bun --entry server.ts --port 3000
 prisma-cli app run api
 ```
 
-## `prisma-cli app deploy [app] --project <id-or-name> --create-project <name> --app <name> --branch <name> --framework <nextjs|nuxt|astro|hono|nestjs|tanstack-start|custom|bun> --entry <path> --http-port <port> --region <region> --env <name=value|file> --db --no-db --prod`
+## `prisma-cli app deploy [app] --project <id-or-name> --create-project <name> --app <name> --branch <name> --framework <nextjs|nuxt|astro|hono|nestjs|tanstack-start|custom|bun> --entry <path> --http-port <port> --region <region> --env <name=value|file> --db --no-db --prod --no-promote`
 
 Purpose:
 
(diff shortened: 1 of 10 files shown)


kristof-siket:
Good catch — fixed in 2417940.

With `--no-promote`, `setKnownLiveDeployment` was persisting the un-promoted candidate (`deployResult.deployment.id`) as the known-live deployment. It now caches the actually-live id — `deployResult.app.liveDeploymentId` (the previous deployment that keeps serving) when not promoted, the new deployment when promoted — and skips the write entirely when the app has no live deployment yet (first deploy with `--no-promote`).

Added a controller test asserting the candidate is **not** cached as live (the store returns the previous live id after a `--no-promote` deploy). `tsc --noEmit`, `biome check`, and the app suite (101 tests) all pass.

kristof-siket:
@coderabbit review

kristof-siket:
@coderabbit approve

PR #110, opened 2026-07-03
feat(cli): add branch remove for preview branches

AmanVarshney01:
Preview branch cleanup previously required the Console. Per the team breakout: removal yes, creation no. Branch creation stays implicit (git-push automation and `app deploy`); there is deliberately no `branch create`.

```bash
$ prisma-cli branch remove feat-login --confirm br_123
✔ Removing branch...
  The branch was removed from the platform. Local Git branches are untouched.

$ prisma-cli branch remove production --confirm br_456
✘ Branch is protected [BRANCH_PROTECTED]

$ prisma-cli branch remove preview --confirm br_123
✘ Branch still has live resources [BRANCH_NOT_EMPTY]
Fix: Rerun with --cascade to remove the branch's apps and databases with it, or remove them individually first.

$ prisma-cli branch remove preview --confirm br_123 --cascade
✔ Removing branch...
  The branch was removed from the platform. Local Git branches are untouched.
  Removed apps: my-app
  Removed databases: preview-db
```

Design notes
- Explicit positional target (branch id or git name in the resolved project) plus the exact-id `--confirm` convention; `--yes` never satisfies it.
- Maps the platform's own guarantees to structured codes: `BRANCH_PROTECTED` (422: production/default branch) and `BRANCH_NOT_EMPTY` (409: live apps or databases remain). Plain removal never touches member resources. Production/default is also refused client-side before the confirm prompt.
- Removal is a platform soft-delete via `DELETE /v1/branches/{branchId}`; local Git branches are untouched.
- Recovery commands render through the detected package runner; new codes (`BRANCH_NOT_FOUND`, `BRANCH_PROTECTED`, `BRANCH_NOT_EMPTY`, `BRANCH_CASCADE_INCOMPLETE`) registered in error-conventions.md.

Review follow-ups (this PR, per @luanvdw)
- `app remove --branch <name>`: the `BRANCH_NOT_EMPTY` recovery hint used to point at a flag that did not exist. The flag is real now, so an app can be removed from a branch that is not checked out locally (or from a directory with no git repo at all).
- `branch remove --cascade`: explicit opt-in teardown for non-production branches. Guardrails: the protection check (production or default role) runs before any member resource is touched, so `--cascade` never widens what can be removed; the exact-id `--confirm` is still required. The CLI removes the branch's apps, then its databases, then the branch, and lists the blast radius in the output and in JSON `result.removed`. A partial failure stops immediately and reports exactly what was already removed (`BRANCH_CASCADE_INCOMPLETE`).
- `BRANCH_NOT_EMPTY` next steps now lead with the `--cascade` rerun, then the per-resource path.

Testing: 9 branch-remove integration tests (confirmation gate, git-name removal, protected, not-empty with cascade-first next steps, not-found, explicit `--project`, cascade success with blast radius, cascade refused on production) plus an `app remove --branch` controller test; full suite 601 green. Also live-verified end to end against a real workspace: real 409 to `BRANCH_NOT_EMPTY` mapping, `app remove --branch` from a git-less directory, cascade removing app + database + branch with the blast radius listed, production refusal with `--cascade`, wrong `--confirm` rejected.

### DIFF
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -865,7 +865,7 @@ prisma-cli branch list
 prisma-cli branch list --json
 ```
 
-## `prisma-cli branch remove <branch> --project <id-or-name> --confirm <branch-id>`
+## `prisma-cli branch remove <branch> --project <id-or-name> --confirm <branch-id> --cascade`
 
 Purpose:
 
@@ -876,9 +876,11 @@ Behavior:
 - requires auth and resolved project context; accepts `--project <id-or-name>` as an explicit fallback
 - resolves `<branch>` by exact Branch id or exact git name within the resolved project
 - requires `--confirm <branch-id>` where the value exactly matches the resolved Branch id; `--yes` does not satisfy this confirmation
-- production Branches and the project's default Branch are protected: removal fails with `BRANCH_PROTECTED`
-- a Branch that still has live Apps or databases fails with `BRANCH_NOT_EMPTY`; branch removal never deletes member resources, so remove the Branch's apps and databases first
-- removal is a platform soft-delete: the Branch disappears from `branch list`, and the platform owns retention of soft-deleted Branches
+- production Branches and the project's default Branch are protected: removal fails with `BRANCH_PROTECTED`, and `--cascade` never widens that; the protection check runs before any member resource is touched
+- without `--cascade`, a Branch that still has live Apps or databases fails with `BRANCH_NOT_EMPTY`; plain removal never deletes member resources, and the recovery steps offer both the cascade rerun and individual `app remove --branch` / `database` cleanup
+- with `--cascade`, the CLI removes the Branch's Apps, then its databases, then the Branch itself; the result lists every removed resource so the blast radius is explicit, in human output and in `result.removed` for `--json`
+- cascade is client-orchestrated because the platform's branch delete refuses non-empty Branches; a mid-cascade failure stops immediately and fails with `BRANCH_CASCADE_INCOMPLETE`, whose `meta` lists what was already removed (removed resources are not restored, and the Branch itself remains)
+- removal is a platform soft-delete: the Branch disappears from `branch list`, and the platform owns retention of soft-deleted Branches; cascaded member resources are removed through their own APIs
 - Branch creation stays implicit (git-push automation and `app deploy`); there is deliberately no `branch create`
 - never touches local Git branches
 - fails with `BRANCH_NOT_FOUND` when no Branch matches
@@ -887,6 +889,7 @@ Examples:
 
 ```bash
 prisma-cli branch remove feat-login --confirm br_123
+prisma-cli branch remove feat-login --confirm br_123 --cascade
 prisma-cli branch remove br_123 --confirm br_123 --json
 ```
 
@@ -1854,16 +1857,18 @@ prisma-cli app rollback
 prisma-cli app rollback --app hello-world --to dep_123
 ```
 
-## `prisma-cli app remove [app] --app <name> -y --yes`
+## `prisma-cli app remove [app] --app <name> --branch <name> -y --yes`
 
 Purpose:
 
-- remove the app from the current branch
+- remove the app from the resolved branch
 
 Behavior:
 
 - requires auth and project context
-- resolves the selected app
+- resolves the branch it reads like the other management commands: explicit `--branch` honored as-is, then the active Git branch when it exists in the project, then the project's default branch
+- `--branch <name>` makes branch cleanup possible for branches that are not checked out locally, such as a teammate's preview branch
+- resolves the selected app within that branch
 - requires confirmation unless `-y` or `--yes` is passed
 - clears local selected app state when the removed app was selected
 
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -865,6 +865,31 @@ prisma-cli branch list
 prisma-cli branch list --json
 ```
 
+## `prisma-cli branch remove <branch> --project <id-or-name> --confirm <branch-id>`
+
+Purpose:
+
+- remove a preview Branch from the resolved project
+
+Behavior:
+
+- requires auth and resolved project context; accepts `--project <id-or-name>` as an explicit fallback
+- resolves `<branch>` by exact Branch id or exact git name within the resolved project
+- requires `--confirm <branch-id>` where the value exactly matches the resolved Branch id; `--yes` does not satisfy this confirmation
+- production Branches and the project's default Branch are protected: removal fails with `BRANCH_PROTECTED`
+- a Branch that still has live Apps or databases fails with `BRANCH_NOT_EMPTY`; branch removal never deletes member resources, so remove the Branch's apps and databases first
+- removal is a platform soft-delete: the Branch disappears from `branch list`, and the platform owns retention of soft-deleted Branches
+- Branch creation stays implicit (git-push automation and `app deploy`); there is deliberately no `branch create`
+- never touches local Git branches
+- fails with `BRANCH_NOT_FOUND` when no Branch matches
+
+Examples:
+
+```bash
+prisma-cli branch remove feat-login --confirm br_123
+prisma-cli branch remove br_123 --confirm br_123 --json
+```
+
 ## `prisma-cli database list --project <id-or-name> --branch <git-name>`
 
 Purpose:
--- a/docs/product/error-conventions.md
+++ b/docs/product/error-conventions.md
@@ -184,6 +184,7 @@ These codes are the minimum stable set for the MVP:
 - `BRANCH_NOT_FOUND`
 - `BRANCH_PROTECTED`
 - `BRANCH_NOT_EMPTY`
+- `BRANCH_CASCADE_INCOMPLETE`
 - `COMPUTE_CONFIG_INVALID`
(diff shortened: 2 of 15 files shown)


ankur-arch:
@AmanVarshney01 here's some feedback on the PR ( in collab with GPT ;)) 

## Maintainability and readability feedback

We should make the branch removal flow easier to read, extend, and safely change over time. The current implementation works, but some parts rely on inline types, repeated string values, and controller-level branching that will become harder to maintain as more branch lifecycle rules are added.

### 1. Name the branch removal result type

We should avoid defining complex union return types directly inside method signatures.

Instead of:

```ts
removeBranch(
  branchId: string,
):
  | { outcome: "removed"; branch: BranchRecord }
  | { outcome: "not-found" }
  | { outcome: "protected" }
  | { outcome: "not-empty" } {
```

Do this:

```ts
type BranchRemoveResult =
  | { outcome: "removed"; branch: BranchRecord }
  | { outcome: "not-found" }
  | { outcome: "protected" }
  | { outcome: "not-empty" };

removeBranch(branchId: BranchId): BranchRemoveResult {
```

This helps because the method signature becomes easier to scan, and the result now has a clear domain name. It also gives us one place to update the result shape if branch removal gains more outcomes later.

---

### 2. Avoid raw strings for important domain states

We should avoid repeating important values like `"removed"`, `"not-found"`, `"protected"`, `"not-empty"`, `"production"`, and `"preview"` directly in multiple places.

Instead of:

```ts
if (branch.role === "production") {
  return { outcome: "protected" };
}
```

Do this:

```ts
const BranchRole = {
  Production: "production",
  Preview: "preview",
} as const;

const BranchRemoveOutcome = {
  Removed: "removed",
  NotFound: "not-found",
  Protected: "protected",
  NotEmpty: "not-empty",
} as const;

if (branch.role === BranchRole.Production) {
  return { outcome: BranchRemoveOutcome.Protected };
}
```

This helps because these values are part of the product behavior, not incidental strings. Centralizing them makes typos less likely and makes it easier to see the supported branch states and removal outcomes.

---

### 3. Extract branch removal rules into small helper functions

We should keep the removal method focused on the flow, not the details of every rule.

Instead of:

```ts
const hasDatabases = (this.data.databases ?? []).some(
  (database) => database.branchId === branchId,
);

const hasDeployments = this.data.deployments.some(
  (deployment) =>
    deployment.projectId === branch.projectId &&
    deployment.branch === branch.name,
);

if (hasDatabases || hasDeployments) {
  return { outcome: "not-empty" };
}
```

Do this:

```ts
function hasLiveBranchResources(data: MockApiData, branch: BranchRecord): boolean {
  return hasBranchDatabases(data, branch.id) || hasBranchDeployments(data, branch);
}

function hasBranchDatabases(data: MockApiData, branchId: BranchId): boolean {
  return (data.databases ?? []).some((database) => database.branchId === branchId);
}

function hasBranchDeployments(data: MockApiData, branch: BranchRecord): boolean {
  return data.deployments.some(
    (deployment) =>
      deployment.projectId === branch.projectId &&
      deployment.branch === branch.name,
  );
}
```

This helps because the main method can read like the product rule: “find branch, reject protected branch, reject branch with live resources, remove branch.” The details are still there, but they are behind names that explain intent.

---

### 4. Keep `runBranchRemove` as orchestration only

We should reduce the amount of decision-making inside `runBranchRemove`. It currently handles auth, project resolution, branch lookup, confirmation, real-vs-fixture deletion, error mapping, and response shaping.

Instead of growing one large controller function, do this:

```ts
export async function runBranchRemove(
  context: CommandContext,
  branchRef: string,
  flags: BranchRemoveFlags,
): Promise<CommandSuccess<BranchRemoveResult>> {
  const removeContext = await resolveBranchRemoveContext(context, flags);
  const branch = resolveBranchForRemoval(branchRef, removeContext.branches, removeContext.target);

  assertBranchCanBeRemoved(branch);
  requireBranchRemoveConfirmation({
    id: branch.id,
    confirm: flags.confirm,
    formatCommand: removeContext.formatCommand,
  });

  await removeBranch(removeContext, branch);

  return createBranchRemoveSuccess(removeContext, branch);
}
```

This helps because each part has one clear job. When something changes later, such as project resolution, confirmation behavior, API handling, or response formatting, we can update that specific part without re-reading the entire command flow.

---

### 5. Use a single mapper from removal outcome to CLI error

We should avoid scattering outcome checks across the controller.

Instead of:

```ts
const removed = context.api.removeBranch(branch.id);

if (removed.outcome === "protected") {
  throw branchProtectedError(branch.gitName);
}

if (removed.outcome === "not-empty") {
  throw branchNotEmptyError(branch.gitName, formatCommand);
}
```

Do this:

```ts
const result = context.api.removeBranch(branch.id);
assertBranchRemoveSucceeded(result, branch, formatCommand);
```

```ts
function assertBranchRemoveSucceeded(
  result: BranchRemoveResult,
  branch: RawBranchRecord,
  formatCommand: PrismaCliPackageCommandFormatter,
): void {
  switch (result.outcome) {
    case "removed":
      return;

    case "protected":
      throw branchProtectedError(branch.gitName);

    case "not-empty":
      throw branchNotEmptyError(branch.gitName, formatCommand);

    case "not-found":
      throw branchNotFoundError(branch.gitName);
  }
}
```

This helps because all outcome-to-error behavior lives in one place. It also makes it easier to check that every possible outcome is handled.

---

### 6. Use clearer names for destructive command inputs

We should make destructive command inputs very explicit.

Instead of:

```ts
branchRef: string
confirm?: string
```

Do this:

```ts
branchRef: BranchRef
confirmBranchId?: BranchId
```

This helps because `branchRef` and `confirmBranchId` are not the same thing. The branch reference can be a branch id or git name, but confirmation must be the exact branch id. Naming that distinction directly makes the safety rule easier to understand.

---

### 7. Move branch removal policy into one place

We should separate the branch removal rules from the mechanics of deleting the branch.

Instead of spreading policy across the mock API, controller, and API response handling, do this:

```ts
function getBranchRemoveBlocker(
  branch: BranchRecord,
  resources: BranchResources,
): BranchRemoveBlocker | null {
  if (branch.role === BranchRole.Production) {
    return "protected";
  }

  if (resources.databases.length > 0 || resources.deployments.length > 0) {
    return "not-empty";
  }

  return null;
}
```

This helps because the product rule becomes reusable and testable on its own: production branches cannot be removed, and branches with live apps or databases cannot be removed. Both the mock path and real path should follow the same rule shape where possible.

---

### 8. Prefer readable condition helpers over nested conditionals

We should avoid nested ternaries and complex inline conditionals in command code.

Instead of:

```ts
const live = providerLiveDeploymentId
  ? deployment.deployment.id === providerLiveDeploymentId
  : knownLiveDeploymentId
    ? deployment.deployment.id === knownLiveDeploymentId
    : deployment.deployment.live;
```

Do this:

```ts
function resolveDeploymentLiveState(options: {
  deploymentId: string;
  defaultLiveState: boolean;
  providerLiveDeploymentId?: string;
  knownLiveDeploymentId?: string;
}): boolean {
  if (options.providerLiveDeploymentId) {
    return options.deploymentId === options.providerLiveDeploymentId;
  }

  if (options.knownLiveDeploymentId) {
    return options.deploymentId === options.knownLiveDeploymentId;
  }

  return options.defaultLiveState;
}
```

This helps because the reader does not need to mentally parse the condition tree. The helper name explains the intent, and the branching becomes easier to test.

---

### Overall recommendation

We should keep the current behavior, but make the implementation more explicit:

* name important result types
* centralize branch roles and removal outcomes
* extract branch removal rules into small helpers
* keep the controller focused on orchestration
* map domain outcomes to CLI errors in one place
* make destructive command inputs clearly named

This will reduce cognitive complexity and make the branch removal flow safer to extend as more branch lifecycle behavior is added.

luanvdw (commented):
Thanks @AmanVarshney01

If I tried this correctly, the `BRANCH_NOT_EMPTY` recovery experience currently sends the user into a dead end. The flow:

1. User runs `prisma-cli branch remove preview --confirm br_123`
2. CLI correctly refuses because the branch still has live resources
3. CLI suggests removing the child app with `prisma-cli app remove --app <name> --branch preview`
4. That command fails, because `app remove` does not currently support `--branch`

So the user is blocked, follows the CLI’s next step, and gets `unknown option '--branch'`.  Can we add `--branch` support to `app remove` rather than only changing the message, because branch cleanup naturally needs “remove this app from this branch, then retry branch removal.”

Let me know if I miss-understood the flow.

Then lastly, (as a follow up, maybe as a separate PR) I think we should support an explicit cascade path for non-production branches. Blocking by default is reasonable, but users should not have to manually tear down every child resource forever. Something like:

`prisma-cli branch remove feat-login --confirm br_123 --cascade`

The important guardrail is that cascade should only be allowed when the resolved branch is not `role=production`, and it should stay refused for production/default branches. Ideally the cascade response also makes the blast radius clear by listing the apps/databases that will be removed, so this is explicit destructive intent rather than a hidden side effect of confirming the branch id.

wdyt?

wmadden-electric:
Closing as stale: this adds `branch remove` to the commander shell, whose registration tree, controllers and mock API were all deleted in #139. The capability — deleting preview branches from the CLI — did not make the engine port (the `branch` group is list-only, matching what shipped before this PR) and is now recorded in `.drive/projects/prisma-cli-v8/deferred.md`. If wanted, it returns as an engine command with the current consent rules (`--confirm <id>`), which did not exist when this was written.

PR #111, opened 2026-07-03
fix(cli): list Create a new Project first in the setup picker

AmanVarshney01:
With many projects, reaching **Create a new Project** in the interactive setup picker meant paging past the entire list. It's now the first option; Cancel stays last. Applies to both `project link` and `app deploy` first-time setup (shared picker).

Before:
```
? Which Project should this directory use?
  ❯ Acme Dashboard
    Billing API
    ... (N more)
    Create a new Project
    Cancel
```

After:
```
? Which Project should this directory use?
  ❯ Create a new Project
    Acme Dashboard
    Billing API
    ...
    Cancel
```

Note the intentional side effect: plain Enter now starts the create flow instead of linking the first listed project. Creating still requires typing a name, so there's no silent resource creation; existing projects are one arrow-key away.

Testing: spec examples updated, six picker-driving tests adjusted for the new order, full suite 591 green.

### DIFF
--- a/docs/product/command-spec.md
+++ b/docs/product/command-spec.md
@@ -1299,12 +1299,13 @@ Behavior:
 
   ```text
   ? Which Project should this directory use?
-    ❯ Acme Dashboard
+    ❯ Create a new Project
+      Acme Dashboard
       Billing API
-      Create a new Project
       Cancel
   ```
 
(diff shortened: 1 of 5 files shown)


kristof-siket (approved):
@AmanVarshney01 💯

luanvdw (approved):

PR #113, opened 2026-07-06
feat(cli): add github command group for workspace GitHub connections

AmanVarshney01:
## Why

Agent-friendly counterpart to the Console work in pdp-control-plane#4396/#4402/#4406/#4407: agents and headless sessions need to drive the GitHub connection flow without a browser. The critical piece is connecting an org that already has the Prisma GitHub App installed — the only path GitHub itself cannot round-trip (no callback fires when nothing changes), now exposed by pdp-control-plane#4408.

## Commands

```bash
prisma-cli github list              # connected + connectable accounts for the active workspace
prisma-cli github connect <account> # connect an already-installed account, by login or installation id
prisma-cli github install           # mint the GitHub App install URL for a brand-new account
```

- **Non-interactive by design**: `connect` without an account fails with `GITHUB_ACCOUNT_REQUIRED`, carrying the connectable options in `error.meta.connectable` and as runnable next steps — an agent picks and reruns, no prompt-blocking. Unknown accounts fail the same way with `GITHUB_ACCOUNT_NOT_FOUND`.
- **Authorization is platform-enforced** (same rule as the Console): member of a workspace the installation is already connected to. A GitHub-side uninstall surfaces as `GITHUB_CONNECT_FAILED` after the platform self-heals its stale rows.
- JSON envelopes, error codes registered in error-conventions.md, spec section added to command-spec.md, runner-formatted next steps.

## Draft status: one dependency

Real mode calls the two new Management API endpoints from pdp-control-plane#4408 (open). The published `@prisma/management-api-sdk` does not carry those paths yet, so they go through a single documented, fully-typed seam in `lib/github/provider.ts` — delete-marked for when the SDK regen lands after #4408 merges. Everything else (`/v1/scm-installations`, `install-intents`) uses the typed SDK client directly.

Undraft once #4408 is merged and the SDK is bumped (seam removal is a follow-up commit here).

## Testing

6 fixture-mode CLI tests (`tests/github.test.ts`): list with per-account dedupe (newest installation wins), connect by login and by numeric id, structured `GITHUB_ACCOUNT_REQUIRED` with machine-readable options, `GITHUB_ACCOUNT_NOT_FOUND` with the same, install URL. Full suite 615 green; typecheck clean. Real mode needs a live check against #4408 once deployed.


wmadden-electric:
Closing as stale: every file this touches — the commander `git` command group, `mock-api.ts`, `command-meta.ts`, the old controllers — was deleted when the Commander shell was retired (#139). The current CLI ships `git connect` / `git disconnect` on the engine. The larger idea here, a `github` group for workspace-level GitHub connections beyond one repo, is not implemented and is now recorded in `.drive/projects/prisma-cli-v8/deferred.md`; if it is still wanted it starts fresh as an engine command, not a rebase of this diff.

PR #117, opened 2026-07-13
feat(cli): run app dev candidates through execa

AmanVarshney01:
Companion to [project-compute#106](https://github.com/prisma/project-compute/pull/106) (approved): the SDK's spawn layer moved to execa there; this migrates the one remaining Windows-broken spawn site in the CLI, the `app dev` launcher ladder in `lib/app/local-dev.ts`.

## Why

The ladder spawned candidates with raw `spawn` and no shell. On Windows, `npx` is a `.cmd` shim, which raw `spawn` cannot launch (only `.exe`/`.com` resolve), so the npx rung always failed with a false ENOENT and Next.js dev degraded to bunx-or-bust. There was even a hand-rolled `next.cmd` special case for the local-bin rung; execa (via cross-spawn, PATHEXT) makes that unnecessary, so it's gone.

execa is already a CLI dependency (`controllers/init.ts`, `controllers/agent.ts`), so no new ecosystem choice.

## Survey of remaining spawn sites (checked all of `packages/`)

- `lib/app/local-dev.ts`: migrated here.
- `shell/update-check.ts`: spawns `process.execPath` (an absolute path to the running Node binary) detached and unref'd. No PATH or shim resolution involved, no output handling, and execa adds nothing but risk around `unref` semantics. Left as is, deliberately.
- `adapters/git.ts`, `lib/git/local-status.ts`: `execFile("git", ...)`; `git.exe` resolves fine on Windows. Left as is.

## Behavior preserved

- Inherited stdio (the dev server streams straight to the terminal).
- ENOENT falls through to the next candidate; ladder exhaustion still throws the friendly install message.
- Exit code and termination signal are reported, and abort still surfaces as SIGTERM so the controller's cancel mapping (`AbortError`) is unchanged.
- Child env passed exactly (`extendEnv: false`).
- One deliberate improvement: a non-ENOENT spawn failure (e.g. EACCES) now surfaces with execa's readable message instead of a raw errno throw.
- Dropped the unused `spawnImpl` injection seam; nothing consumed it (tests mock `runLocalApp` itself).

## Testing

- Full suite: 45 files / 624 tests pass; typecheck and build clean.
- Live smoke of the migrated path: a real `bun --watch` app spawns, streams output, and terminates cleanly on abort (`SIGTERM` reported); an all-ENOENT ladder (empty PATH) produces the friendly message.

### DIFF
--- a/packages/cli/src/lib/app/local-dev.ts
+++ b/packages/cli/src/lib/app/local-dev.ts
@@ -270,9 +270,7 @@ async function runWithFallback(
 /**
  * Runs one candidate to completion with inherited stdio, returning its exit
  * information, or "unavailable" when the binary does not exist so the ladder
- * can try the next candidate. execa spawns through cross-spawn, so Windows
- * `.cmd` shims (npx, `node_modules/.bin` entries) resolve via PATHEXT instead
- * of failing with a false ENOENT.
+ * can try the next candidate.
  */
 async function spawnCommand(
   candidate: CommandCandidate,
@@ -288,10 +286,9 @@ async function spawnCommand(
       signal: NodeJS.Signals | null;
     }
 > {
-  // The caller passes the full runtime env; extending over process.env again
-  // would let parent vars leak past a deliberate omission.
   const result = await execa(candidate.command, candidate.args, {
     cwd: options.cwd,
+    // The env is fully constructed by the caller; do not re-merge process.env.
     env: options.env,
     extendEnv: false,
     cancelSignal: options.signal,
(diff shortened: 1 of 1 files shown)


wmadden-electric:
Closing as superseded by the platform port. This PR improves `app run`'s dev-server spawning in `src/lib/app/local-dev.ts` — both the command and that file were deleted when the Commander shell was retired (#139). `app run` was ruled out of the new CLI entirely (operator, 2026-08-11): running a local dev server is Composer's `dev`. Nothing from this change is still needed.

PR #124, opened 2026-07-20
fix(cli): don't assume newest deployment is live in promote resolution

kristof-siket:
## Why

`resolveCurrentLiveDeploymentId` fell back to `deployments[0]` (the **newest** deployment) when no authoritative live signal was available (`app.liveDeploymentId`, the provider `live` flag, or local known-live state):

```ts
return deployments[0]?.id ?? null;
```

During `app promote`, this made a **freshly created candidate** look like it was *already live*, so promote emitted "The selected deployment is already live for this app." and **skipped the endpoint rebind**. An app whose live pointer had been cleared (e.g. its live deployment was stopped) therefore never got re-bound — every subsequent deploy hit the same false "already live" and left the app unreachable.

This was the second half of a production incident on 2026-07-20: after the live deployments of build-runner/github-webhook were stopped (clearing `latestDeploymentId`), each CI redeploy's promote step read the null pointer, hit this fallback, and skipped the rebind — turning a transient stop into a ~1h outage.

## What

Return `null` instead of guessing the newest. "Unknown live deployment" now means **not already-live**, so promote proceeds and rebinds the endpoint.

## Tests

- New: `promote rebinds instead of assuming the newest deployment is live when there is no authoritative live signal`.
- Full `app-controller` suite green.

### DIFF
--- a/packages/cli/src/controllers/app.ts
+++ b/packages/cli/src/controllers/app.ts
@@ -3097,6 +3097,14 @@ function requireDeploymentForApp(
   });
 }
 
+/**
+ * Resolves the app's live deployment from the app pointer, the provider's live
+ * flag, then locally cached state.
+ *
+ * @returns the live deployment id, or null when no authoritative signal exists.
+ * Callers must treat null as "not known to be live" rather than assuming the
+ * newest deployment is live.
+ */
 async function resolveCurrentLiveDeploymentId(
   context: CommandContext,
   projectId: string,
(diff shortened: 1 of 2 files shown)


kristof-siket:
@coderabbit approve

kristof-siket:
@coderabbitai approve

kristof-siket:
@coderabbitai review

PR #129, opened 2026-08-09
Add @prisma/cli-engine — one library owns argv to exit code — proven end to end on auth whoami

wmadden-electric:
This is what writing a command looks like when this PR's engine owns everything between `argv` and the exit code:

```ts
import { defineCommand } from "@prisma/cli-engine";
import { ok } from "@prisma/cli-engine/protocol";

export const authWhoamiCommand = defineCommand({
  help: {
    summary: "Show the authenticated user and accessible workspace",
    examples: ["auth whoami", "auth whoami --json"],
  },
  handler: async (_args, ctx) => {
    const state = await readAuthState(ctx.env, ctx.signal);
    return ok(
      ctx.present(
        { data: state },
        {
          human: () => [
            { kind: "summary", tone: "info", text: "Showing the current authenticated identity." },
            { kind: "fields", rows: fieldRows(state) },
          ],
          stdout: () => fieldRows(state).map((r) => `${r.label}: ${r.value}`),
          next: () => (state.authenticated ? [] : [SIGN_IN]),
        },
      ),
    );
  },
});
```

The command declares words and a handler; the engine does the rest. The same definition produces the human card on stderr with pipeable payload lines on stdout, and — with `--json`, or automatically when stdout is not a TTY — a machine stream ending in one result envelope:

```json
{"kind":"result","envelope":{"ok":true,"commandId":"auth.whoami","result":{"authenticated":false,...},"exitCode":0,"diagnostics":[],"nextActions":[{"kind":"run-command","label":"Sign in","command":"prisma-cli auth login"}]},"commandId":"auth.whoami","timestamp":"..."}
```

## The decision

Prisma ships three CLIs today (`prisma-next`, `prisma-composer`, `@prisma/cli`) with three config files and three unrelated help/output/error dialects. We are consolidating them into one `prisma` binary, and **this PR makes the foundational move: a single library, `@prisma/cli-engine`, owns every CLI mechanic — parsing, preconditions, context, presentation, envelopes, exit codes, prompts, signals, config loading — so that a product package contributes only command definitions, handlers, and words.** To prove the design is real rather than aspirational, the PR also ports one real command (`auth whoami`) end to end through an unpublished bin.

## How the engine models a command

A command **settles like a promise**. It either **completes** — possibly unsuccessfully: findings travel as typed diagnostics inside a completed envelope with a documented exit code — or it **errors**, aborting with a structured error the engine renders. This one distinction drives everything downstream: envelopes, exit codes (0 completed / 1 bug / 2 errored / 3 user cancel / 4–99 documented / 130,143 signals), and rendering.

Around that core, the engine owns:

- **Formats and channels.** `--format human|json` (auto-json on non-TTY stdout). Human mode: presentation to stderr, machine-usable payload lines to stdout — human mode is pipe-clean. Json mode: one frame per line on stdout, exactly one terminal envelope.
- **Preconditions (`needs`).** Credentials, config sections, optional dependencies, interactivity — checked before the handler loads, failing early with engine-phrased structured errors.
- **Prompts** (defaults accepted by `--yes`; `consent` structurally undefaultable), **events** (a fixed vocabulary with a per-command `data` extension), **log levels** (`--quiet` and `--verbose` are aliases on one severity scale).
- **Config.** `prisma.config.ts` is discovered and evaluated once; `defineConfig` stamps a version marker; a Prisma 7 config fails early with a typed migration error. A broken config only fails commands that need config.
- **Process boundaries.** The engine never touches process globals: streams, env, TTY flags, signals, and `exit` arrive injected via a `Runtime`. First Ctrl-C aborts gracefully; a second force-exits. This is also what makes it fully testable.

Products plug in as a **`CommandFamily`** — the unit of contribution and ownership: commands by name, at most one config section, a docs base URL. The shell owns the tree (paths, groups); a command may only require its own family's config section, enforced at construction.

## Proving it

- `@prisma/cli-engine/testing` ships `createTestCli`: argv-in/bytes-out harness testing with injected credentials, config, clock, prompt answers, and signals. The package's 145 tests byte-assert human output, json frames, exit codes, prompt semantics, signal handling, and config loading — including permanent compile-verified type-tests (a stale `@ts-expect-error` fails the build).
- `prisma-v8` (unpublished, tsx-run: `pnpm prisma-v8 auth whoami`) mounts the ported command over the existing token storage. Behavioral differences from `prisma-cli auth whoami` are enumerated, not discovered: [the divergence record](.drive/projects/prisma-cli-v8/assets/engine/whoami-parity-divergences.md).

## What this PR does not do

Nothing ships to users: the engine is unpublished, the bin is not wired into any package's `bin`, and the published `@prisma/cli` build is untouched. The other ~60 platform commands, auth-library extraction, Composer and ORM adoption, and the release pipeline are later slices ([plan](.drive/projects/prisma-cli-v8/plan.md)).

## Alternatives considered

- **Clipanion (powers two of the three CLIs today) vs stricli** for the internal parser: stricli 1.3.0, exact-pinned and fully hidden — nothing stricli-shaped appears in the public surface ([comparison record](.drive/projects/prisma-cli-v8/assets/engine/stricli-vs-clipanion.md)).
- **Bundling the parser vs an ordinary dependency**: ordinary exact-pinned dependency — bundling is unusual for a library and blinds security audit.
- **Lazily imported handler modules** (`handler: () => import(...)`): rejected — handlers are direct function references; the indirection bought startup weight savings we didn't need at the cost of a circular typing pattern.
- **`Result`-wrapped prompt returns**: rejected — prompts return plain values or throw; the engine settles failures uniformly (cancel → exit 3).
- **Schema-library validation (e.g. Arktype) for construction checks**: deferred — a heavy dependency against minimal validation needs; revisit if the checks grow.

The full interface record, including every amendment made during implementation, is the normative draft at [.drive/projects/prisma-cli-v8/assets/engine/engine-interface-draft.ts](.drive/projects/prisma-cli-v8/assets/engine/engine-interface-draft.ts); requirements live in [docs/architecture/cli-engine-requirements.md](docs/architecture/cli-engine-requirements.md) (merged via #128).


packages/cli-engine/src/config-loader.ts
```diff
@@ -0,0 +1,121 @@
...
+  };
+}
+
+export function stampConfigMarker<T extends Record<string, unknown>>(
+  config: T,
+): T & { readonly $prismaConfig: number } {
+  return Object.freeze({ ...config, $prismaConfig: PRISMA_CONFIG_VERSION });
+}
+
+export async function loadConfigImpl(cwd: string): Promise<LoadedConfig> {
```
wmadden: Why does `config-loader.ts` export something called `loadConfigImpl()`? There doesn't appear to be an interface this implements

packages/cli-engine/src/config-loader.ts
```diff
@@ -0,0 +1,121 @@
...
+
+const MARKER_KEY = "$prismaConfig";
+
+function isMarked(value: unknown): value is Record<string, unknown> {
+  return (
+    typeof value === "object" &&
+    value !== null &&
+    typeof (value as Record<string, unknown>)[MARKER_KEY] === "number"
+  );
+}
```
wmadden: What does this do? What does "marked" mean? This is unreadable

packages/cli-engine/src/config-loader.ts
```diff
@@ -0,0 +1,121 @@
...
+function fileLevelConfig(diagnostic: Diagnostic): LoadedConfig {
+  return { sections: {}, diagnostics: [{ section: null, diagnostic }] };
+}
+
+function missingMarkerDiagnostic(path: string): Diagnostic {
+  return {
+    code: "CLI.CONFIG_MISSING_MARKER",
+    severity: "error",
+    summary:
+      "prisma.config.ts is not a Prisma v8 config: its default export does not carry the defineConfig version marker.",
```
wmadden: This is not a meaningful user-facing error

packages/cli-engine/src/config-loader.ts
```diff
@@ -0,0 +1,121 @@
...
+      {
+        kind: "user-choice",
+        label: "Fix the error in the file, then run the command again.",
+      },
+    ],
+    where: { path },
+  };
+}
+
+export function stampConfigMarker<T extends Record<string, unknown>>(
```
wmadden: This function looks like a violation of encapsulation, is it? Normally a branded type keeps the stamping _internal_. Possibly this is not branding? Possibly it should not be called "stamp"

packages/cli-engine/src/index.ts
wmadden, line 1: Follow the pattern in prisma/prisma and composer of separating package exports/ from src/. Do not mix them like this.

packages/cli-engine/src/core.ts
wmadden, line 1: 1k LOC is too long. Split this file.

packages/cli-engine/src/core.ts
```diff
@@ -0,0 +1,958 @@
...
+    }
+  | {
+      readonly kind: "artifact";
+      readonly path: string;
+      readonly description?: string;
+      readonly data?: unknown;
+    };
+
+// —————————————————————————————————————————————————————————————————————
+// §2 Outcomes and presented results
```
wmadden: What are these section marker? This is code, not prose

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+// —————————————————————————————————————————————————————————————————————
+
+function constructionError(message: string): Error {
+  return new Error(`@prisma/cli-engine: ${message}`);
+}
+
+const CAMEL_CASE = /^[a-z][a-zA-Z0-9]*$/;
+const INTEGER_LIKE = /^\d+$/;
+
+function validateFlags(path: string, def: AnyCommand): void {
```
wmadden: This looks like something that should be implemented with eg Arktype

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+    const code = Number(key);
+    if (!Number.isInteger(code) || code < 4 || code > 99) {
+      throw constructionError(
+        `command '${path}' documents exit code ${key}; documented codes must be integers in 4-99`,
+      );
+    }
+  }
+}
+
+interface MountEntry {
```
wmadden: Not clear what a "mount entry" or "mount node" is

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+interface MountNode {
+  readonly commands: Map<string, MountEntry>;
+  readonly children: Map<string, MountNode>;
+}
+
+function emptyNode(): MountNode {
+  return { commands: new Map(), children: new Map() };
+}
+
+function insertMount(
```
wmadden: Is a mount a noun or verb?

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+function productOf(
+  spec: EngineSpec,
+  def: AnyCommand,
+): ProductManifest | undefined {
+  return spec.products.find((manifest) =>
+    Object.values(manifest.commands).includes(def),
+  );
+}
+
+function validateProductSection(
```
wmadden: wtf is a "product section"?

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+  spec: EngineSpec,
+  path: string,
+  def: AnyCommand,
+): void {
+  const section = def.needs?.config;
+  if (section === undefined) {
+    return;
+  }
+  const product = productOf(spec, def);
```
wmadden: Don't love this

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+    return;
+  }
+  if (product.configSection !== section) {
+    throw constructionError(
+      `command '${path}' needs the '${section.name}' config section, which is not its product's declared section`,
+    );
+  }
+}
+
+function buildMountTree(spec: EngineSpec): MountNode {
```
wmadden: What is a mount tree? You mean a command tree?

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+  const now = options?.now ?? (() => new Date());
+  return {
+    execute: (argv, runtime, hooks) =>
+      executeRun(spec, root, argv, runtime, { ...hooks }, now),
+  };
+}
+
+// —————————————————————————————————————————————————————————————————————
+// A single run
+// —————————————————————————————————————————————————————————————————————
```
wmadden: wtf is this?

packages/cli-engine/src/execution.ts
```diff
@@ -0,0 +1,2180 @@
...
+
+function sniffFormat(argv: readonly string[], runtime: Runtime): Format {
```
wmadden: This is too cryptic

packages/cli-engine/src/execution.ts
wmadden, line 1: This file has way too many concerns in it and IT'S 2k LOC LONG! SPLIT IT

packages/cli-engine/src/protocol.ts
```diff
@@ -0,0 +1,208 @@
...
+  assertOk(): T {
+    return this.value;
+  }
+
+  assertNotOk(): never {
+    throw new Error("Expected NotOk result but got Ok");
+  }
+}
+
+class NotOkImpl<F> implements NotOk<F> {
```
wmadden, line 174: Did you reimplement this? Or copy it from where it was already impleemnted in one of the original repos?

packages/cli/src/shell/command-runner.ts
```diff
@@ -49,10 +50,7 @@ function toCliError(error: unknown, runtime: CliRuntime): CliError | null {
     });
   }
 
-  if (
-    error instanceof Error &&
-    error.message.startsWith("PRISMA_SERVICE_TOKEN is set but empty")
-  ) {
+  if (error instanceof EmptyServiceTokenError) {
```
wmadden: Don't use instanceof for this. Use a discriminator on the error

packages/cli/src/v8/auth/whoami.ts
```diff
@@ -0,0 +1,111 @@
...
+  return [
+    { label: "status", value: "signed in" },
+    ...(user ? [{ label: "user", value: user }] : []),
+    ...(state.provider
+      ? [{ label: "provider", value: providerLabel(state.provider) }]
+      : []),
+    ...(state.workspace?.name
+      ? [{ label: "workspace", value: state.workspace.name }]
+      : []),
+  ];
```
wmadden: This is unreadable

packages/cli/src/v8/main.ts
```diff
@@ -0,0 +1,145 @@
...
+    groups: {
+      auth: { brief: "Manage local authentication for the CLI" },
+    },
+    commands: {
+      "auth whoami": authWhoamiCommand,
+    },
+  });
+}
+
+export interface SignalProcess {
```
wmadden: This interface looks like it belongs in the cli engine runtime

wmadden (changes requested):

wmadden-electric:
Response to the review round (all landed, `0e088ae..94524f4`, verified by an independent pass):

| Comment | Resolution |
| --- | --- |
| execution.ts 2.2k LOC — split it | Split into 11 modules under `src/execution/`, largest 421 LOC (c4b091e) |
| core.ts 1k LOC — split it | Dissolved into 10 modules under `src/definition/` (c4b091e) |
| Separate exports from src | Entry modules now `src/exports/{index,protocol}.ts` per the composer/prisma convention (c4b091e); shared tsconfig extracted to the `@repo/tsconfig` workspace package (4c3a085) |
| Section banner comments | All removed; module boundaries carry the structure |
| `loadConfigImpl` | → `loadConfig` (no interface existed; the name was noise) |
| "marked" | → `hasVersionMarker` |
| `stamp` / encapsulation | Helper deleted; `defineConfig` owns the marker internally, key stays module-local |
| mount entry/node/tree | → `CommandTreeEntry` / `CommandTreeNode` / `buildCommandTree`; "mount" survives only as a verb |
| "product section" / `productOf` / "product is not an entity" | `ProductManifest` → **`CommandFamily`** (ruled vocabulary), `products:` → `commandFamilies:`, `productOf` → `commandFamilyOf`; "product" and "manifest" expelled from engine code, draft, and project docs |
| Cryptic format sniff | Extracted `explicitFormat()` with the pre-parse rationale stated |
| whoami unreadable rows | Nested spread-ternaries replaced with plain conditionals |
| Meaningless config error | Missing-marker diagnostic rewritten in user-facing words |
| Arktype for validation | **Deferred by ruling** — heavy dependency, minimal validation needs (recorded in the divergence doc) |
| `instanceof` in command-runner | `EmptyServiceTokenError` gains a `code` discriminator; both shells match on it |
| `ProcessLike` belongs in the engine | Moved as `HostProcess` in the engine's runtime surface |
| Reimplemented or copied? | Adapted port of the prisma/prisma donor sources (`0-foundation/utils`, `1-core/errors/control.ts`) with the settled adjustments |

Also in the range: `ctx.env` (closes the R4 hole — handlers no longer read `process.env`), json-mode `--help` keeps stdout frame-clean, the prompt stdin reader no longer hangs the real bin, `--version` works (exit 0), the shipped `buildCli()` tree is constructed in tests, and the npm rollout plan is recorded at `.drive/projects/prisma-cli-v8/assets/rollout-plan.md`.

Verification: engine 144 tests, cli 720, typecheck green, lint exit 0 in CI conditions; dist `.d.ts` delta confirmed to be exactly the rename + `ctx.env` + `HostProcess`.

packages/cli-engine/src/command-family.ts
```diff
@@ -0,0 +1,22 @@
+import type { AnyCommand } from "./commands";
+import type { ConfigSection } from "./config-section";
+
+/**
+ * The unit of contribution and ownership a package exports for CLI
+ * purposes: its config section (declared once — a family-level fact)
+ * and its commands by NAME. A command whose needs.config token is not
+ * its command family's section is a construction error. The shell owns
+ * the tree; a command family owns its section (R12).
```
wmadden: NO TRANSIENT PROJECT IDS

packages/cli-engine/src/commands.ts
```diff
@@ -0,0 +1,239 @@
...
+export interface HelpSpec {
+  /** One line, imperative, shown in listings. */
+  readonly summary: string;
+  readonly description?: string;
+  /**
+   * Invocations WITHOUT the binary name: at help render time every
+   * `{bin}` is substituted with createCli's `name`; an example
+   * containing no `{bin}` gets the name prepended.
+   */
+  readonly examples?: readonly string[];
```
wmadden, line 40: Don't define types with conditional properties. The define methods can accept conditional options but should normalize to a consistent structure. The examples array can be empty. The description can be `string | undefined`. 

Apply this to all types in this PR.

packages/cli-engine/src/envelopes.ts
wmadden, line 1: "Envelopes" isn't a great grouping - it's a type, not a logical grouping. Completed/Errored envelopes belong with the handler definition. Stream events with whatever they're relevant to.

packages/cli-engine/src/events.ts
```diff
@@ -0,0 +1,86 @@
+import type { NextAction } from "../protocol";
+
+/**
+ * The commentary severity scale; also the log-level axis. Distinct from
+ * Diagnostic severity: 'verbose' grades commentary, which never enters
+ * the envelope.
+ */
+export type Severity = "error" | "warn" | "info" | "verbose";
+export type LogLevel = Severity;
```
wmadden: Why do we have two names for the same thing?

packages/cli-engine/src/presentation.ts
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * What a completed command's handler returns inside `ok(...)`: the
+ * outcome plus the presentation the active format already materialized.
+ * Built exclusively by ctx.present — the brand makes hand-construction
+ * a type error. Human rendering writes the blocks, next-action lines,
+ * and diagnostics to stderr and the `stdout` lines to stdout — human
+ * mode is pipe-clean (operator ruling, 2026-08-09).
+ */
+export interface PresentedResult<T> {
+  readonly [PRESENTED]: true;
```
wmadden, line 35: Why is the symbol necessary?

packages/cli-engine/src/testing.ts
wmadden, line 1: Test code does not go mixed in with production code!

packages/cli-engine/src/execution/invocation.ts
wmadden, line 1: Why is this named invocation.ts instead of after its primary export: engine.ts?

packages/cli-engine/src/execution/run.ts
```diff
@@ -0,0 +1,421 @@
...
+  spec: EngineSpec,
+  options?: { readonly now?: () => Date },
+): Engine {
+  const root = buildRoutes(spec, buildCommandTree(spec), "", executeMounted);
+  const now = options?.now ?? (() => new Date());
+  return {
+    execute: (argv, runtime, hooks) =>
+      executeRun(spec, root, argv, runtime, { ...hooks }, now),
+  };
+}
```
wmadden: This is engine construction, not "running". WHy is it in a file called run.ts?

packages/cli-engine/src/execution/run.ts
```diff
@@ -0,0 +1,421 @@
...
+  readonly groups: Readonly<
+    Record<string, { readonly brief: string; readonly description?: string }>
+  >;
+  readonly commands: MountedTree;
+}): Cli {
+  const engine = buildEngine(spec);
+  return {
+    run: (argv, runtime) => engine.execute(argv, runtime, {}),
+  };
+}
```
wmadden: This is CLI construction, not "running".

packages/cli-engine/src/execution/run.ts
```diff
@@ -0,0 +1,421 @@
...
+
+async function executeRun(
```
wmadden: What is "execute run"? Run already means execute

packages/cli-engine/src/execution/run.ts
wmadden, line 1: A lot of the code here looks like it belongs on an Engine class instead of lumped together as "run".ts.

Implement an actual class eg EngineImpl while retaining its interface Engine.

packages/cli-engine/src/execution/run.ts
```diff
@@ -0,0 +1,421 @@
...
+    timestamp: invocation.now().toISOString(),
+  });
+  return 0;
+}
+
+/** Maps stricli's own settlement (parse/route failures, framework bugs)
+ *  onto the engine protocol when the pipeline never settled. A failure
+ *  addressed at a server command renders to stderr — a foreign client on
+ *  the other end of stdout must never receive an engine envelope. */
+function settleUnhandled(
```
wmadden: Why are there all these "settleX()" functions if there's a separate file called settlement.ts?

packages/cli-engine/src/args.ts
```diff
@@ -0,0 +1,173 @@
+/**
+ * Single-character alias, enforced at the type level: `Char<'q'>` is
+ * 'q'; `Char<'ab'>` is never.
+ */
+export type Char<S extends string> = S extends `${string}${infer Rest}`
```
wmadden, line 5: Does this need to be exported?

packages/cli-engine/src/args.ts
```diff
@@ -0,0 +1,173 @@
...
+  ? Rest extends ""
+    ? S
+    : never
+  : never;
+
+export const FLAG: unique symbol = Symbol("prisma.cli-engine.flag");
```
wmadden: What's with all the symbol brands? Isn't that overkill?

packages/cli-engine/src/command-family.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export interface CommandFamily {
+  readonly configSection?: ConfigSection<unknown>;
+  readonly commands: Readonly<Record<string, AnyCommand>>;
+  /**
+   * The family's documentation base URL. The engine derives each
+   * diagnostic's docs link from base + code.
+   */
+  readonly docsBaseUrl?: string;
```
wmadden, line 24: Why is this optional?

packages/cli-engine/src/commands.ts
```diff
@@ -0,0 +1,239 @@
...
+ * `kind`. Definitions load their handler's import graph at startup;
+ * R9's remaining force is that handler bodies defer heavy work to
+ * execution time.
+ */
+
+/** The help SPI: words only — the engine formats. */
+export interface HelpSpec {
+  /** One line, imperative, shown in listings. */
+  readonly summary: string;
+  readonly description?: string;
```
wmadden, line 34: Why is this optional?

packages/cli-engine/src/commands.ts
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * Invocations WITHOUT the binary name: at help render time every
+   * `{bin}` is substituted with createCli's `name`; an example
+   * containing no `{bin}` gets the name prepended.
+   */
+  readonly examples?: readonly string[];
```
wmadden, line 40: Why isthis optional.

wmadden (changes requested):

packages/cli-engine/src/config-loader.ts
wmadden-electric: Renamed to loadConfig — there was no interface; the Impl suffix was noise (48496dd).

packages/cli-engine/src/config-loader.ts
wmadden-electric: Renamed to hasVersionMarker; the check names the $prismaConfig version marker explicitly (48496dd).

packages/cli-engine/src/config-loader.ts
wmadden-electric: Rewritten as a user-facing message naming the file, what is wrong, and what to do (48496dd).

packages/cli-engine/src/config-loader.ts
wmadden-electric: The exported helper is gone — defineConfig attaches the marker itself and the key stays module-local (48496dd).

packages/cli-engine/src/index.ts
wmadden-electric: Entry modules now live in src/exports/ per the composer/prisma convention; shared tsconfig extracted to @repo/tsconfig (c4b091e, 4c3a085).

packages/cli-engine/src/core.ts
wmadden-electric: core.ts is gone — dissolved into flat single-concept modules in src/ (c4b091e, 2ec5c04).

packages/cli-engine/src/core.ts
wmadden-electric: All banner comments removed; module boundaries carry the structure now (c4b091e).

packages/cli-engine/src/execution.ts
wmadden-electric: Deferred by your ruling — heavy dependency for minimal validation needs; recorded in the divergence doc.

packages/cli-engine/src/execution.ts
wmadden-electric: Renamed CommandTreeEntry / CommandTreeNode (48496dd).

packages/cli-engine/src/execution.ts
wmadden-electric: Noun uses are gone; "mount" survives only as the verb for placing a command at a path (48496dd).

packages/cli-engine/src/execution.ts
wmadden-electric: "Product" is expelled — the entity is CommandFamily and the check reads "its command family's section" (c4b091e).

packages/cli-engine/src/execution.ts
wmadden-electric: productOf is now commandFamilyOf over the ruled CommandFamily vocabulary (c4b091e).

packages/cli-engine/src/execution.ts
wmadden-electric: Yes — renamed buildCommandTree / CommandTreeNode (48496dd).

packages/cli-engine/src/execution.ts
wmadden-electric: Extracted as explicitFormat() with the pre-parse rationale stated in code (48496dd).

packages/cli-engine/src/execution.ts
wmadden-electric: Rewritten with named helpers in shared-flags.ts (48496dd).

packages/cli-engine/src/execution.ts
wmadden-electric: Split into single-concern modules under src/execution/, largest ~420 LOC (c4b091e).

packages/cli-engine/src/protocol.ts
wmadden-electric: Adapted port of the prisma/prisma donor sources (0-foundation/utils, 1-core/errors/control.ts) with the settled adjustments — not a reimplementation.

packages/cli/src/shell/command-runner.ts
wmadden-electric: EmptyServiceTokenError carries a code discriminator now; both shells match on it (7807753).

packages/cli/src/v8/auth/whoami.ts
wmadden-electric: The nested spread-ternaries are plain conditionals now (48496dd).

packages/cli/src/v8/main.ts
wmadden-electric: Moved into the engine as HostProcess in the runtime surface (7807753).

packages/cli-engine/src/command-family.ts
wmadden-electric: Done — every R-number citation in cli-engine and cli/src/v8 comments is replaced with the rule stated in plain words (71848dc). The architecture doc keeps the numbering.

packages/cli-engine/src/commands.ts
wmadden-electric: Done in 71848dc: definitions normalize help, args, needs, and exitCodes to always-present fields; CommandFamily gets defineCommandFamily; Diagnostic and CliStructuredError always carry nextActions (empty when none); PresentedResult materializes every presentation field. The define* inputs stay optional — the constructors normalize.

packages/cli-engine/src/envelopes.ts
wmadden-electric: Done — envelopes.ts is deleted: CompletedEnvelope/ErroredEnvelope now sit beside the handler definitions in commands.ts, and StreamEvent/StreamMeta live in events.ts (71848dc).

packages/cli-engine/src/events.ts
wmadden-electric: Done — the LogLevel alias is deleted; Severity is the one name for the scale, and the flag stays --log-level (71848dc).

packages/cli-engine/src/presentation.ts
wmadden-electric: Kept, as ruled: the PRESENTED symbol is what makes ctx.present the only way to build a PresentedResult — a handler cannot hand-construct one. The arg-spec symbols, which had no such job, are gone (71848dc); the type-tests pin the forgery claim.

packages/cli-engine/src/testing.ts
wmadden-electric: Done — the harness moved out of the main entry onto its own public subpath, @prisma/cli-engine/testing (entry src/exports/testing.ts, mirroring composer's ./testing convention); all tests import it from there (fecf8b9).

packages/cli-engine/src/execution/invocation.ts
wmadden-electric: Done — renamed to execution/engine.ts, whose primary exports are Engine/EngineSpec and the EngineImpl class (d6edb90).

packages/cli-engine/src/execution/run.ts
wmadden-electric: Done — run.ts is deleted. buildEngine now lives in execution/engine.ts and constructs an EngineImpl (d6edb90).

packages/cli-engine/src/execution/run.ts
wmadden-electric: Done — createCli moved to src/cli.ts beside the Cli interface (d6edb90).

packages/cli-engine/src/execution/run.ts
wmadden-electric: Done — no function named executeRun survives; the pipeline is EngineImpl.execute (d6edb90).

packages/cli-engine/src/execution/run.ts
wmadden-electric: Done — there is now an EngineImpl class implementing the Engine interface; the execute pipeline (execute, executeMounted, executeServer) is its methods, and run.ts is gone (d6edb90).

packages/cli-engine/src/execution/run.ts
wmadden-electric: Done — every settle* function, including settleVersion and settleUnhandled from run.ts, now lives in settlement.ts (d6edb90).

packages/cli-engine/src/args.ts
wmadden-electric: No — the exported symbol is gone. The spec types now carry a never-assigned optional phantom property instead, which keeps declaration emit working without exporting anything (71848dc).

packages/cli-engine/src/args.ts
wmadden-electric: Agreed — FLAG/POSITIONAL are replaced with plain phantom properties (__flag? / __positional?), never set at runtime. Only PRESENTED keeps its brand, because it enforces ctx.present as the sole constructor (71848dc).

packages/cli-engine/src/command-family.ts
wmadden-electric: Done — CommandFamily's fields are now required (typed T | undefined); defineCommandFamily takes the optional spec and normalizes (71848dc).

packages/cli-engine/src/commands.ts
wmadden-electric: Done — the field is required on the definition; defineCommand accepts it optionally and normalizes (71848dc).

packages/cli-engine/src/commands.ts
wmadden-electric: Done — required on the definition, normalized by the constructor; only the define* input keeps it optional (71848dc).

wmadden (dismissed):

.drive/projects/prisma-cli-v8/assets/briefs/s1-handover.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Open items for the operator (do not resolve unilaterally)
+
+1. D3 interpretation: in human non-quiet mode the engine renders human Blocks only; the materialized `stdout` presentation lines are written only under `--quiet` (they would duplicate the blocks). The draft honors materialization exactly; the rendering-rule reading is unruled. Revisit during D6 whoami parity.
+2. D4 interpretation: remediation events render as nextActions at settlement in human mode (not live in the transcript); json streams them live as frames.
+3. Diagnostic severity stays `error|warn|info`; the trim to two awaits the ADR 239 amendment (project slice S4).
+4. D5: a file-level config diagnostic (unmarked/unreadable `prisma.config.ts`) fails EVERY command, including ones with no config need — the draft's `LoadedConfig` comment says so and won over the 1a foundation design, which says the opposite. Needs an operator ruling before S3.
+5. D5: diagnostics returned on a SUCCESSFUL `SectionValidation` (warnings) are currently dropped; the draft doesn't say where they go.
+
+Put all of these on the slice PR's parity-divergence list for Will's review.
```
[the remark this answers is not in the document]
wmadden-electric: The brief under .drive/projects/prisma-cli-v8/assets/briefs/ is a historical record of what was handed over at the time, not live documentation. The settled rulings live in whoami-parity-divergences.md, which is the document a later agent reads.

.drive/projects/prisma-cli-v8/assets/engine/whoami-parity-divergences.md
```diff
@@ -0,0 +1,226 @@
...
+stdout goes. v8 auto-selects json when stdout is not a TTY (draft
+header: human on a TTY stdout, json otherwise). Piping
+`prisma-v8 auth whoami` therefore produces the json stream where the
+current CLI would produce the human card on stderr.
+
+## 3. Human output: channel and rendering
+
+Current CLI (signed out):
+
+```
```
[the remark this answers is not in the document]
wmadden-electric: .drive is excluded from this repo's lint config (biome.jsonc `!.drive`) and markdownlint is not run here. The file is a project decision record, so churning it for a linter the repo does not use adds noise.

.drive/projects/prisma-cli-v8/assets/rollout-plan.md
```diff
@@ -0,0 +1,65 @@
...
+   operator owns both repos and sequences it). From then on
+   `prisma-next` carries v8 rc builds under a `next` tag. Rationale for
+   waiting until S5: `prisma-next`'s installed base is ORM early
+   adopters, and the unified CLI cannot serve them before the ORM
+   family exists.
+4. **`prisma7` ships (owned by the ORM team, timing open).** The
+   v7-and-under train moves to the `prisma7` name, freeing the bare
+   `prisma` name. This repo configures OIDC trusted publishing for
+   `prisma` and publishes `prisma@8.0.0-rc1` under the **`next`/`rc`
+   dist-tag** (the S7 pipeline's artifact).
```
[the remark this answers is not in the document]
wmadden-electric: Out of scope for this slice. Publish machinery is S7. The rollout plan records the intended shape; the exact dist-tag commands get written and tested when the publish workflow is built.

.github/workflows/preview-cli-package.yml
```diff
@@ -44,6 +44,12 @@ jobs:
             --sha '${{ github.event.pull_request.head.sha }}' \
             --pr-number '${{ github.event.pull_request.number }}' >> "$GITHUB_OUTPUT"
 
+      # Build every workspace package up front via the turbo graph so
+      # workspace:* dependencies (the CLI's @prisma/cli-engine) have
+      # their dist output before the test and build steps below.
+      - name: Build workspace packages
+        run: pnpm build
+
```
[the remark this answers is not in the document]
wmadden-electric: The per-package build steps predate this PR and carry the `cli_build` / `compute_build` step ids the later status reporting reads. The up-front turbo build was added so workspace:* dependencies have dist output before the test steps; the repeated tsdown runs cost a few seconds and keep the status wiring intact.

package.json
```diff
@@ -6,19 +6,23 @@
   },
   "packageManager": "pnpm@11.6.0",
   "scripts": {
+    "build": "turbo run build",
     "build:cli": "pnpm --filter @prisma/cli build",
```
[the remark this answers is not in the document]
wmadden-electric: `pnpm --filter @prisma/cli build` skips the turbo ^build graph, so running it on a clean checkout builds the CLI against a missing or stale engine dist. Real footgun, one-line fix. Fixing this in the PR.

packages/cli-engine/src/commands.ts
```diff
@@ -0,0 +1,371 @@
...
+    args: Args<TFlags, Record<never, PositionalSpec<unknown>>>,
+    io: {
+      readonly stdin: InputStream;
+      readonly stdout: OutputStream;
+      readonly stderr: OutputStream;
+      readonly signal: AbortSignal;
+      readonly cwd: string;
+      readonly config: TConfig;
+    },
+  ) => Promise<number>;
```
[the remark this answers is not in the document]
wmadden-electric: CommandContext exposes `ctx.env` precisely so handlers never read process.env. Server handlers only get `cwd`, so that rule has no way to hold for them. Runtime.env is already at the dispatch site. Fixing this in the PR.

packages/cli-engine/src/execution/engine.ts
```diff
@@ -0,0 +1,384 @@
...
+        stdout: runtime.stdout,
+        stderr: runtime.stderr,
+        signal: invocation.signal,
+        cwd: runtime.cwd,
+        config: needsOutcome.config,
+      });
+      state.settledExitCode = exitCode;
+    } catch (cause) {
+      settleThrown(invocation, cause);
+    }
```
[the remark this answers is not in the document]
wmadden-electric: Result commands are guarded by undocumentedExitCode; server commands assign the handler's return value straight to settledExitCode. NaN exits 0 and 300 exits 44, so a handler defect becomes a wrong process status with no report. Fixing this in the PR.

packages/cli-engine/src/execution/engine.ts
```diff
@@ -0,0 +1,384 @@
...
+  for (const argument of argv) {
+    if (argument === "--") {
+      return false;
+    }
+    if (argument === "--version") {
+      return true;
+    }
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
wmadden-electric: Confirmed: execute() intercepts `--version` before stricli parses, but RESERVED_FLAG_NAMES has no `version`, so a command may declare a `version` flag that can never be reached and construction says nothing. Fixing this in the PR.

packages/cli-engine/src/execution/needs.ts
```diff
@@ -0,0 +1,246 @@
...
+    if (fileLevel.length > 0) {
+      return needsErrored(
+        structuredErrorFromDiagnostic(fileLevel[0].diagnostic),
+        fileLevel.slice(1).map((entry) => entry.diagnostic),
+      );
+    }
+    return validateConfigSection(needs.config, invocation);
+  }
+  return { kind: "ok", config: undefined };
+}
```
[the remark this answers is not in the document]
wmadden-electric: noExcessiveCognitiveComplexity is enabled in this repo's biome.jsonc and reports 16 against a limit of 15. The extraction is mechanical and keeps the check order. Fixing this in the PR.

packages/cli-engine/src/execution/needs.ts
```diff
@@ -0,0 +1,246 @@
...
+
+export function dependencyResolvable(specifier: string, cwd: string): boolean {
+  try {
+    createRequire(join(cwd, "__cli_engine_probe__.js")).resolve(specifier);
+    return true;
+  } catch {
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
wmadden-electric: Runtime.cwd is host-supplied and nothing in the interface says it must be absolute. createRequire rejects a relative path, the bare catch swallows it, and every dependency check then reports CLI.MISSING_DEPENDENCY. Fixing this in the PR.

packages/cli-engine/src/execution/prompts.ts
```diff
@@ -0,0 +1,237 @@
...
+    confirm: async (question, opts) => {
+      const fallback = opts?.default;
+      if (state.yes || !state.interactive) {
+        if (fallback === undefined) {
+          throw promptUnanswerable(question, state);
+        }
+        return fallback;
+      }
+      const hint =
+        fallback === undefined ? "(y/n)" : fallback ? "(Y/n)" : "(y/N)";
```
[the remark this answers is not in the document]
wmadden-electric: noNestedTernary is enabled in this repo's biome.jsonc. One-line fix with identical output. Fixing this in the PR.

packages/cli-engine/src/execution/rendering.ts
```diff
@@ -0,0 +1,183 @@
...
+  const newline = text.indexOf("\n");
+  return (newline === -1 ? text : text.slice(0, newline)).trim();
+}
+
+const STEP_OUTCOME_SYMBOL: Readonly<Record<string, string>> = {
+  ok: "✔",
+  failed: "✖",
+  skipped: "↷",
+  warning: "⚠",
+};
```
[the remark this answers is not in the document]
wmadden-electric: Both are Record<string, string>, so a new outcome or tone silently renders the text 'undefined' to the terminal. DIAGNOSTIC_SYMBOL in the same file already uses the union and catches this at compile time. Fixing this in the PR.

packages/cli-engine/src/execution/rendering.ts
```diff
@@ -0,0 +1,183 @@
...
+ *  when the diagnostic does not carry its own — a per-raise docsUrl wins. */
+export function withDocsUrl(
+  state: RunState,
+  diagnostic: Diagnostic,
+): Diagnostic {
+  if (diagnostic.docsUrl !== undefined || state.docsBaseUrl === undefined) {
+    return diagnostic;
+  }
+  return { ...diagnostic, docsUrl: `${state.docsBaseUrl}${diagnostic.code}` };
+}
```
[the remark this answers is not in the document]
wmadden-electric: Confirmed: nothing validates docsBaseUrl at command-tree construction, and withDocsUrl concatenates directly. A family registering a base without a trailing slash ships a broken docs link in both human output and the JSON envelope. Fixing this in the PR.

packages/cli-engine/src/execution/settlement.ts
```diff
@@ -0,0 +1,295 @@
...
+      severity: "error",
+      summary: firstLine(
+        cause instanceof Error ? cause.message : String(cause),
+      ),
+      nextActions: [],
+    },
+    diagnostics: [],
+    nextActions: [],
+  });
+}
```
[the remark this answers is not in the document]
wmadden-electric: Real debuggability gap, but where an internal-error stack surfaces is a design decision about the diagnostic shape, which is settled for S1 (typed nextActions, no free-form fields). Tracked for the diagnostics work. Tracked as TML-3190: https://linear.app/prisma-company/issue/TML-3190/cli-engine-preserve-the-stack-of-an-internal-error-for-debugging

packages/cli-engine/src/execution/stricli-adapter.ts
```diff
@@ -0,0 +1,294 @@
...
+  values: readonly (string | undefined)[],
+) => Promise<void>;
+
+function parseNumberInput(input: string): number {
+  const value = Number(input);
+  if (!Number.isFinite(value)) {
+    throw new Error(`expected a number, received '${input}'`);
+  }
+  return value;
+}
```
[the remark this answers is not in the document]
wmadden-electric: Number('') is 0 and 0 is finite, so `--count=` parses to 0 instead of failing with a usage error. Fixing this in the PR.

packages/cli-engine/src/exports/index.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * The v8 engine surface: everything a command-family package imports
+ * for CLI purposes — definition constructors, flag/positional builders,
+ * the context/envelope/runtime types, and the createCli entry point.
+ * The test harness lives on the ./testing subpath.
+ *
+ * Normative source: .drive/projects/prisma-cli-v8/assets/engine/
+ * engine-interface-draft.ts (v8).
+ */
```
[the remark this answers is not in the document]
wmadden-electric: The two entrypoints are deliberately different surfaces: ./protocol is the shapes that cross package and process boundaries and drags no engine code, so other processes can depend on it alone. Duplicating those symbols on the barrel would give every type two import paths.

packages/cli-engine/src/exports/testing.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * The ./testing subpath: the in-memory test harness — the same engine
+ * over in-memory streams. Implementation lives in ../testing.ts.
+ */
+export { createTestCli, type TestCli } from "../testing";
```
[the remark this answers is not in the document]
wmadden-electric: Agreed that the warning is a false positive here: src/exports/{index,protocol,testing}.ts are the tsdown entrypoints and a re-export hub is their intended shape. A scoped override keeps the Lint check free of standing warnings. Fixing this in the PR.

packages/cli-engine/src/presentation.ts
```diff
@@ -0,0 +1,107 @@
...
+      readonly text: string;
+    }
+  | {
+      readonly kind: "fields";
+      readonly rows: ReadonlyArray<{
+        label: string;
+        value: string;
+        sensitive?: boolean;
+      }>;
+    }
```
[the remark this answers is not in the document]
wmadden-electric: Every other member of Block is fully readonly; these rows are the exception, and tightening the type after the package ships is a breaking change. Fixing this in the PR.

packages/cli-engine/src/protocol.ts
```diff
@@ -0,0 +1,209 @@
...
+    this.nextActions = options?.nextActions ?? [];
+    this.where = options?.where
+      ? {
+          ...ifDefined("path", options.where.path),
+          ...ifDefined("line", options.where.line),
+        }
+      : undefined;
+    this.meta = options?.meta;
+    this.docsUrl = options?.docsUrl;
+  }
```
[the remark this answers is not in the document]
wmadden-electric: auth whoami already converts a caught EmptyServiceTokenError into a CliStructuredError and drops the original. Additive, non-breaking, and keeps the envelope unchanged. Fixing this in the PR.

packages/cli-engine/src/testing.ts
```diff
@@ -0,0 +1,166 @@
...
+      const abort = opts?.abort;
+      if (abort !== undefined) {
+        if (abort.aborted) {
+          deliverSignal(abort.reason);
+        } else {
+          abort.addEventListener("abort", () => deliverSignal(abort.reason), {
+            once: true,
+          });
+        }
+      }
```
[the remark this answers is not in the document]
wmadden-electric: There is no such window. engine.execute subscribes to runtime.onSignal in its synchronous prefix, before the first await, so the subscription is already in place by the time execute() returns the promise and the harness wires abort delivery.

packages/cli-engine/tests/command-families.test.ts
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  test("human rendering shows the derived docs link", async () => {
+    const result = await familyCli().run(["failing"], {
+      isTty: { stdout: true },
+    });
+
+    expect(result.stderr).toBe(
+      "✖ [TOY.BROKEN] It broke\n" + `  docs: ${BASE}TOY.BROKEN\n`,
+    );
```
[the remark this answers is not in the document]
wmadden-electric: useTemplate is enabled in this repo's biome.jsonc. One-line fix with the same expected string. Fixing this in the PR.

packages/cli-engine/tests/engine.type-test.ts
```diff
@@ -0,0 +1,389 @@
...
+  env: { CI: "1" },
+  isTty: { stdin: true, stdout: true, stderr: true },
+  exit: (code: number): never => {
+    throw new Error(String(code));
+  },
+  onSignal: () => () => {},
+  config: loadedConfig,
+  getCredentials: async () => undefined,
+  packageManager: "pnpm",
+};
```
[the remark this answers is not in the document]
wmadden-electric: runtimeShape exists to pin the Runtime interface as a whole; stdin's type is already checked by the Runtime annotation. Whether the value is a working async iterator is exercised by the behavioural tests, not by this shape.

packages/cli-engine/tests/execution.test.ts
```diff
@@ -0,0 +1,964 @@
...
+              severity: "error",
+              summary: "The config file could not be read",
+              nextActions: [],
+            },
+          },
+        ],
+      },
+      getCredentials: async () => opts.credentials,
+      packageManager: "unknown",
+    };
```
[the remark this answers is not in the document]
wmadden-electric: These three tests deliberately spell out the whole runtime they exercise, which is what makes them readable in isolation. The shared harness for everything else is @prisma/cli-engine/testing.

packages/cli-engine/tests/lifetimes.test.ts
```diff
@@ -0,0 +1,463 @@
...
+  });
+
+  test("a second SIGTERM exits 143", async () => {
+    const { runtime, exited, deliver } = hangingRuntime();
+    void stuckCli().run(["stuck"], runtime);
+
+    deliver("SIGINT");
+    expect(() => deliver("SIGTERM")).toThrow("runtime.exit(143)");
+    expect(exited).toEqual([143]);
+  });
```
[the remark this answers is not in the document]
wmadden-electric: The stuck handler never settles by design, so there is no rejection to catch, and Biome's noFloatingPromises rule is satisfied by the explicit void.

packages/cli-engine/tests/lifetimes.test.ts
```diff
@@ -0,0 +1,463 @@
...
+    handler: async (_args, io) => {
+      let received = "";
+      const decoder = new TextDecoder();
+      for await (const chunk of io.stdin) {
+        received += decoder.decode(chunk, { stream: true });
+      }
+      io.stdout.write(`Content-Length: ${received.length}\r\n\r\n${received}`);
+      return 0;
+    },
+  });
```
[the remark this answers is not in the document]
wmadden-electric: The fixture is a toy in a test file, not shipped example code, and nothing asserts its framing. It exists to prove the engine hands a server command raw stdin and its exit code, which it does.

packages/cli-engine/tests/prompts.test.ts
```diff
@@ -0,0 +1,505 @@
...
+  test("CI suppresses interactivity even on a TTY", async () => {
+    const result = await cliWith(confirmDefaultTrue).run(["probe"], {
+      ...INTERACTIVE,
+      env: { CI: "true" },
+    });
+
+    expect(result.exitCode).toBe(0);
+    expect(result.presented?.data).toEqual({ answer: true });
+    expect(result.stderr).toBe("✔ answer=true\n");
+  });
```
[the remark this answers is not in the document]
wmadden-electric: The rule is presence, not value: defaultInteractive checks runtime.env.CI === undefined, and the existing test pins it. Adding per-value cases would pin a value-sensitive rule the engine deliberately does not have.

packages/cli-engine/tests/protocol.type-test.ts
```diff
@@ -0,0 +1,38 @@
...
+    : false
+  : false;
+
+export const envelopeIsOkFalseAndDiagnostic: MutuallyAssignable<
+  CliErrorEnvelope,
+  { readonly ok: false } & Diagnostic
+> = true;
+
+export const envelopeFromError: CliErrorEnvelope =
+  undefined as unknown as ReturnType<CliStructuredError["toEnvelope"]>;
```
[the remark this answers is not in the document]
wmadden-electric: toEnvelope's return type is annotated CliErrorEnvelope at the method itself, so the class cannot widen it. The assignment here checks the export wiring, which is what it is for.

packages/cli/src/v8/runtime.ts
```diff
@@ -0,0 +1,92 @@
...
+export function makeGetCredentials(
+  env: NodeJS.ProcessEnv,
+): () => Promise<Credentials | undefined> {
+  return async () => {
+    const serviceToken = env[SERVICE_TOKEN_ENV_VAR]?.trim();
+    if (serviceToken) {
+      return { token: serviceToken };
+    }
+    const tokens = await new FileTokenStorage(env).getTokens();
+    return tokens ? { token: tokens.accessToken } : undefined;
```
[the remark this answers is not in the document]
wmadden-electric: readAuthState treats a blank service token as invalid configuration and errors; makeGetCredentials trims it to '' and silently reads the stored token instead. The two auth paths must agree on what a blank variable means. Fixing this in the PR.

packages/cli/tests/v8-bin.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    expect(await getCredentials()).toEqual({ token: "svc_token" });
+  });
+
+  it("reads the stored token even after the run signal aborted", async () => {
+    const controller = new AbortController();
+    controller.abort("SIGINT");
+    const getCredentials = makeGetCredentials({} as NodeJS.ProcessEnv);
+
+    expect(await getCredentials()).toEqual({ token: "stored_token" });
+  });
```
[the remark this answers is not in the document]
wmadden-electric: Confirmed: the AbortController is local and never reaches makeGetCredentials, which takes only env. The test name claims behaviour the assertion does not check. Fixing this in the PR.

turbo.json
```diff
@@ -0,0 +1,26 @@
...
+        "fixtures/**",
+        "package.json",
+        "tsconfig.json",
+        "tsdown.config.ts",
+        "vitest.config.ts"
+      ]
+    },
+    "typecheck": {
+      "dependsOn": ["^build"],
+      "inputs": ["src/**", "tests/**", "package.json", "tsconfig.json"]
```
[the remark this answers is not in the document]
wmadden-electric: Confirmed: the tasks declare explicit inputs, which replace the default discovery, and @repo/tsconfig has no build task to contribute a hash. Editing base.json today leaves every cached build, test, and typecheck result valid. Fixing this in the PR.

package.json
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/commands.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/engine.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/engine.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/needs.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/needs.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/prompts.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/rendering.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/rendering.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/execution/stricli-adapter.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/exports/testing.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/presentation.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/src/protocol.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli-engine/tests/command-families.test.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli/src/v8/runtime.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

packages/cli/tests/v8-bin.test.ts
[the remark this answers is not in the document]
wmadden-electric: On it 👍

turbo.json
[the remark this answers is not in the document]
wmadden-electric: On it 👍

package.json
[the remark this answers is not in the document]
wmadden-electric: Done in 28c1ac5: build:cli now runs `turbo run build --filter=@prisma/cli`, so a clean checkout builds @prisma/cli-engine before @prisma/cli.

packages/cli-engine/src/commands.ts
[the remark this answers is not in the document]
wmadden-electric: Done in d9eb615: the server io now carries env from Runtime.env, covered by a test; the draft interface asset is amended to match.

packages/cli-engine/src/execution/engine.ts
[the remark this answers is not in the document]
wmadden-electric: Done in d9eb615: a server handler's exit code must be an integer in 0-255; anything else settles as CLI.INTERNAL_ERROR with exit 1, covered by tests.

packages/cli-engine/src/execution/engine.ts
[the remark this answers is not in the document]
wmadden-electric: Done in d9eb615: 'version' is now in RESERVED_FLAG_NAMES, so declaring a version flag fails construction, covered by a test.

packages/cli-engine/src/execution/needs.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 9dae410: checkNeeds is split into one helper per need with the check order unchanged; the complexity notice is gone.

packages/cli-engine/src/execution/needs.ts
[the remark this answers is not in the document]
wmadden-electric: Done in d9eb615: the probe path now uses resolve(cwd, ...), and a test covers a relative runtime cwd.

packages/cli-engine/src/execution/prompts.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 9dae410: replaced with an explicit conditional.

packages/cli-engine/src/execution/rendering.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 7e1d122: both symbol maps are keyed by their union types, matching DIAGNOSTIC_SYMBOL.

packages/cli-engine/src/execution/rendering.ts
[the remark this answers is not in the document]
wmadden-electric: Done in d9eb615: the join normalizes the trailing slash, and a family's docsBaseUrl must parse as a URL at construction; both covered by tests.

packages/cli-engine/src/execution/stricli-adapter.ts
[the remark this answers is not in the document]
wmadden-electric: Done in d9eb615: blank or whitespace-only numeric values are usage errors, covered by a test.

packages/cli-engine/src/exports/testing.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 9dae410: noBarrelFile is off for packages/cli-engine/src/exports/** via a scoped biome override and on everywhere else.

packages/cli-engine/src/presentation.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 7e1d122: the fields rows are readonly.

packages/cli-engine/src/protocol.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 7e1d122: the constructor accepts an optional cause and forwards it to Error; the envelope is unchanged, covered by a test.

packages/cli-engine/tests/command-families.test.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 9dae410: one template literal.

packages/cli/src/v8/runtime.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 64568d0: a set-but-blank PRISMA_SERVICE_TOKEN now throws EmptyServiceTokenError, matching readAuthState, instead of reading the stored token; covered by a test.

packages/cli/tests/v8-bin.test.ts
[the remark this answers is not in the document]
wmadden-electric: Done in 64568d0: the unused AbortController is gone, the test is renamed to the stored-token read it performs, and a blank-token test is added.

turbo.json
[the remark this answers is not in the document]
wmadden-electric: Done in 28c1ac5: $TURBO_ROOT$/packages/tsconfig/base.json joins the build, test, and typecheck inputs; a turbo dry run confirms it is hashed.

PR #137, opened 2026-08-11
Brief for the last platform slice: port init, then delete the old CLI

wmadden-electric:
A handover brief so the final platform slice can be picked up by an agent with no context on this project.

## What it covers

Port the two commands still on the old shell — the `init` wizard and `version` — then delete the commander shell and the fixture machinery, cut the shipped binary over, and close the platform port.

## Why it is written this way

It is meant to be actionable, not a summary of the contract. Four things it does that the contract does not:

**It says when to start.** Two pull requests must land first, because both add commands to the tree this slice completes and delete code it removes. Starting early means rebasing a very large deletion across two moving branches.

**It separates the blocked part from the rest.** The shipped binary cannot currently read `prisma.config.ts` — the loader does a plain dynamic import, which only works because everything runs under `tsx` today. Choosing the fix changes what a user must install before the CLI works, so it is the operator's call. Only the binary cutover waits on it; the ports, the deletions and the grammar check do not.

**It sizes the deletion and names the trap.** Roughly 20,000 lines across five directories, of which an unknown part survives as the operation layer the ported commands still call — so the brief asks for the survivors to be enumerated in the pull request. And it calls out the knot: `src/auth/errors.ts` still constructs the old shell's error class, so the auth module depends on the shell it is meant to outlive.

**It flags a defect in the contract.** The paragraph on the binary cutover contradicts itself mid-sentence — "Resolution is pinned… STOP: that is not pinned anywhere" — a previous author catching their own error. The brief points at the question ledger as authoritative rather than leaving the next reader to work that out.

## What is in it

Where the work sits, what to do in what order, what the engine surface actually looks like after the last slice (it changed substantially, so older documents are stale), the verification commands including the build-first trap that has cost time twice, and the process rules: bot identity, explicit staging, pull-request structure, report style.

Documentation only.

### DIFF
--- a/.drive/projects/prisma-cli-v8/assets/briefs/init-and-shell-retirement.md
+++ b/.drive/projects/prisma-cli-v8/assets/briefs/init-and-shell-retirement.md
@@ -11,7 +11,7 @@ Repo `prisma/prisma-cli`. This is the last slice of the platform port. Two thing
 
 When this lands, `prisma-cli` runs entirely on `@prisma/cli-engine` and the old shell is gone.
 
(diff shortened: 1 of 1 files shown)


wmadden-electric:
Folded into #139, which now carries the brief and the work it briefed. The brief's commits merged into that branch unchanged.

PR #141, opened 2026-08-11
docs(spec): redirect tables in command-family metadata

wmadden-electric:
## Grounding example

`migration apply` was renamed to `migrate --to` long ago. Today the ORM CLI answers the old name with a targeted message; on the engine it would get a generic unknown-command error, because a family has no way to declare a retired invocation.

With this capability the family declares it as metadata:

```ts
defineCommandFamily({
  commands,
  redirects: [
    { from: 'migration apply', replacement: '{bin} migrate --to <ref>', reason: 'migration apply was replaced by migrate --to.' },
  ],
})
```

```
$ prisma migration apply
✖ `migration apply` has been replaced
  migration apply was replaced by migrate --to.
  → Use the replacement: prisma migrate --to <ref>
```

## The decision

`CommandFamily` gains an optional `redirects` table, consulted only when an invocation fails to resolve. Entries are metadata, not commands: never in help, never in the grammar tree, never executable. `{bin}` in a replacement is substituted at render time, which is what lets a family name a replacement invocation without hard-coding a binary name (R12) — the same family mounts under `prisma-next` today and `prisma` at cutover. Retired flags on live commands work the same way. A new engine-owned code `CLI.COMMAND_MOVED` carries it, exit 2, with the replacement as a `run-command` next action.

The alternative the ORM port would otherwise be stuck with is registering dead verbs as real commands, which the engine cannot hide — two useless entries in `--help` forever, inherited by the unified tree.

## What this PR is

The implementation spec only — `.drive/projects/prisma-cli-v8/specs/engine-redirect-table.md` — written for handoff to an independent implementer. Includes the surface, matching and substitution rules, four construction-time validations, explicit non-behavior, test obligations, and a note on the files it overlaps with the Composer stream and the package-manager capability.


wmadden-electric:
Folded into #142, which carries this spec and the implementation together.

PR #142, opened 2026-08-11
A renamed command now says what replaced it, instead of failing like a typo

wmadden-electric:
## At a glance

We renamed `database` to `postgres`, with no alias. Here is what a user who types the old name gets today:

```
$ prisma-v8 database list
✖ [CLI.UNKNOWN_COMMAND] No command registered for `database`
```

The same answer they would get for a typo. We know the name was retired and we know what replaced it; we say neither.

This PR lets the command family that owns those commands say so:

```ts
defineCommandFamily({
  commands,
  redirects: [
    {
      from: 'database list',
      replacement: 'postgres list',
      reason: 'database was renamed to postgres.',
    },
  ],
})
```

```
$ prisma-v8 database list
✖ [CLI.COMMAND_MOVED] `database list` has been replaced
  why: database was renamed to postgres.
→ Use the replacement: prisma-v8 postgres list
```

## The decision

A `CommandFamily` gains an optional `redirects` table. Entries are **metadata, not commands**: they never enter the mounted tree or the route map, so they cannot appear in help, cannot be executed, and have no handler. The engine consults the table only after an invocation has already failed to resolve — if nothing matches, the user gets exactly the error they get today.

## What you write, and why

**`from` is the invocation as the user types it** — an absolute path in the shipped tree, the same convention as the mount map's keys. It has to be, because it is matched against what someone typed. It is not relative to anything: a family has no single place in the tree, and this repo's own platform family is scattered across six roots.

**`replacement` carries no binary name.** It is written the way help examples are written and rendered by the same function they use — `{bin}` substituted where present, the binary name prepended where not. That is what lets a family name a runnable replacement without hard-coding a binary, which matters because the same family mounts under `prisma-v8` today and `prisma` at cutover.

**A retired flag on a live command works the same way**, declared as a `(command path, flag)` pair:

```ts
{ from: 'migration status', flag: 'graph', replacement: 'migration graph' }
```

Either kind settles as a new engine-owned code, `CLI.COMMAND_MOVED`, at exit 2, carrying the replacement as a typed `run-command` action — so an agent or a script can act on it without parsing prose.

## What fails at build time

Five defects throw at `createCli`, alongside every other tree defect. Four are mechanical: a retired flag naming a command that does not exist, a retired flag the command actually accepts, a retired flag name that is not camelCase (matching camel-cases what the user typed, so a kebab-cased entry would look plausible and never fire), and two entries claiming the same invocation — checked across families, so two families cannot both claim a path.

The fourth is the interesting one: **a retired verb whose `from` resolves to a live command or a live group.** The group half is easy to miss. Stricli resolves a bare group path to help rather than to a routing failure, so a redirect sitting on a group could never fire — making it a build-time error is the only way that mistake ever becomes visible. It also earns its keep elsewhere: matching walks prefixes of what the user typed to resolve longest-match, which is only safe because any prefix of a mounted path is a live group and a redirect cannot sit there.

## One rule that had to come along

When a run addressed at a server command fails, the engine forces human rendering, because a foreign client owns stdout and must never receive an engine envelope. The redirect settlement is a second exit from that same failure path and inherits it. Without that, a retired flag on a server command would have written a JSON frame into a channel an LSP client is reading — not an edge case, since server commands run non-TTY and non-TTY means json.

## Known limits

**Re-mounting a group silently stops a family's redirects matching.** `from` is a fact about the shell's tree, written in family code, and no check can catch a stale one — a `from` naming something that does not exist is the normal case. Re-grouping means re-reading the redirect table by hand.

**The engine's error codes are still catalogued nowhere**; this makes nineteen undocumented ones. Documenting the set is deferred until the list has settled rather than done piecemeal here.

**A pre-existing json gap is inherited, not introduced.** A command that hands the terminal to a child refuses `--json` thoroughly, but that refusal lives in the mounted-execution path, which a parse failure never reaches — so an unknown flag on such a command already emits a json frame today, and a retired flag now does the same. Guarding only the new settlement would make it disagree with its sibling, so it is left alone and recorded.

## Alternatives

**Register the retired names as real commands.** This is what the ORM port would otherwise be stuck with — it already answers two retired verbs and four retired flags with targeted messages, and that behavior had nowhere to go on the engine. Rejected because the engine cannot hide a mounted command: every dead verb becomes a permanent entry in `--help`, inherited by the unified tree forever.

**Let the shell mount redirects instead of the family.** Removes the re-mount risk entirely, since the retired path would sit in the same file as the live path replacing it. Rejected because a retirement is product knowledge — the shell would have to carry a table of every product's history.

**Let a live command quietly win a collision instead of failing the build.** Rejected because an overlap is never a legitimate steady state: either two repos disagree about whether a name is retired, or someone reintroduced a name and left the redirect behind. Precedence keeps that mistake alive and silent; a build-time error surfaces it once, in CI.


wmadden (approved):

wmadden:
@coderabbitai trigger now

wmadden-electric:
Both review comments are addressed in `8385214`.

**Non-camelCase redirect flag names** — confirmed as a real bug before fixing. With the check removed, a redirect declared `flag: "skip-generate"` constructs without complaint and typing `--skip-generate` yields `CLI.INVALID_ARGUMENTS`; the replacement never appears. `addFlagRedirect` now tests `flag` against the existing `CAMEL_CASE` rule before the table key is built, throwing in the neighbouring checks' voice. It runs first in the function, ahead of the mounted-command lookup, because it is a well-formedness check on the declaration itself that needs no tree and guarantees a key is never built from a bad name. That makes five construction-time validations; the spec and the description are updated to match.

**Cross-family duplicate for flag redirects** — added, alongside a test for the kebab-case rejection above.

PR #145, opened 2026-08-11
Ctrl-C exits 130 because the engine says so, not because a command invented a dead child

wmadden-electric:
Press Ctrl-C during `prisma composer dev` and the CLI should exit 130, the conventional code for "terminated by SIGINT". Before this change it exited 0, because a session command's successful return settled 0 no matter what had happened to the process — so the only way to reach 130 was to hand the child-status bypass a **child result describing a child that never existed**:

```ts
// composer's dev command, before
const INTERRUPTED: ChildResult = { exitCode: null, signal: "SIGINT" };
...
return ok(exitWithChildStatus(INTERRUPTED)); // a fiction, for the arithmetic
```

Now the handler cleans up and returns normally, and the engine settles 130 by itself:

```ts
return ok(undefined);
```

## The decision

**Signal exit codes belong to the engine, sourced from its own record.** The engine already knows which signal was delivered — that is the record-and-replay latch that lets Ctrl-C reach a spawned child natively. That record now decides the exit code for both command kinds, including a handler that caught the abort, cleaned up, and returned successfully. Nothing a handler returns can produce a signal exit code.

Two paths stay verbatim, deliberately: a spawned child's own status, and a server command's protocol conclusion. Neither number was this CLI's to author — a forwarded SIGTERM still settles the child's own 42, and the abort ladder still settles 137 rather than 130.

The invented-child trick is now impossible rather than merely discouraged: `ctx.spawn` marks every child result it produces, and the settlement refuses one that the engine did not mint. It is a single extra condition beside the existing declaration fence, in the same place, taking the same construction-error route.

The engine's exit-code doctrine is amended to say all of this, and the handler-facing docs now tell authors that cleanup never states an exit code.

## Alternatives considered

- **An explicit "canceled" outcome for session handlers.** More machinery for the same result, and it lets a handler claim a cancellation that never happened. The engine's own record cannot lie.
- **Leaving the stand-in in composer.** The exit code was right, but the value was a lie a reader would take at face value, and it would break the moment anything validated that a child result came from a real spawn — which is exactly what this change adds.

### DIFF
--- a/packages/cli-engine/src/execution/settlement.ts
+++ b/packages/cli-engine/src/execution/settlement.ts
@@ -5,7 +5,7 @@ import type {
 } from "../commands";
 import { PRESENTED, type PresentedResult } from "../presentation";
 import { CliStructuredError, type Diagnostic } from "../protocol";
-import type { ChildStatusSettlement } from "../spawn";
+import { type ChildStatusSettlement, isEngineSpawned } from "../spawn";
 import type { EngineSpec, Invocation } from "./engine";
 import {
   firstLine,
@@ -59,7 +59,8 @@ export function settleCompleted(
   }
   const state = invocation.state;
   invocation.hooks.onPresented?.(presented);
-  state.settledExitCode = presented.exitCode;
+  const exitCode = runExitCode(invocation, presented.exitCode);
+  state.settledExitCode = exitCode;
   if (state.format === "json") {
     const envelope: CompletedEnvelope = {
       ok: true,
@@ -68,7 +69,7 @@ export function settleCompleted(
         presented.presentation.json === undefined
           ? presented.data
           : presented.presentation.json,
-      exitCode: presented.exitCode,
+      exitCode,
       diagnostics: presented.diagnostics.map((diagnostic) =>
         withDocsUrl(state, diagnostic),
       ),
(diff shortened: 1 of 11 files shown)


packages/cli-engine/src/execution/settlement.ts
```diff
@@ -106,10 +107,27 @@ export function settleErrored(
   });
 }
 
-/** Signal exit codes: 130 SIGINT, 143 SIGTERM. The engine's own
- *  controller only ever aborts with a signal name. */
-function signalExitCode(reason: unknown): number {
-  return reason === "SIGTERM" ? 143 : 130;
+/** The conventional code for a delivered signal: 128 + its number, so
+ *  130 for SIGINT and 143 for SIGTERM. */
+function signalExitCode(signal: "SIGINT" | "SIGTERM" | undefined): number {
```
wmadden: Why is `undefined` permitted?

wmadden (approved):

packages/cli-engine/src/execution/settlement.ts
wmadden-electric: It should not be. Neither caller could pass `undefined` — `runExitCode` narrows it away first, and the run signal is aborted from exactly one place, which records the signal before aborting. Worse, the fallback would have mapped that impossible `undefined` to 130, claiming SIGINT for an abort that carried no signal.

The parameter now takes `"SIGINT" | "SIGTERM"`. The one site still holding the optional field, `settleAborted`, decides explicitly: no recorded signal settles cancelled (3) rather than inventing one.

Fixed in 3d18c6e.

PR #147, opened 2026-08-11
A stray space in a retired command path no longer stops it matching

wmadden-electric:
## At a glance

A redirect table declared like this builds without complaint:

```ts
redirects: [
  { from: 'migration  apply', replacement: 'migrate --to <ref>' },
]
```

Two spaces. Typing `prisma migration apply` still gets `CLI.UNKNOWN_COMMAND`; the replacement never appears. Nothing fails, nothing warns, and the entry sits in the table forever matching nothing.

Now it just works — the path is read the way it was obviously meant.

## Why it silently failed

`from` was stored verbatim as the table key, but lookup takes the leading argv tokens and rejoins them with **single** spaces. A key holding a doubled space is one the lookup can never construct.

None of the five checks in #142 caught it, because they all ask the same question — *does this collide with something real?* A path with a stray space collides with nothing, which is the correct answer for a redirect: a `from` naming something that doesn't exist is the entire point of the feature. Nothing asked whether the path was one an invocation could produce.

## The fix

`from` is normalized where a `RedirectSpec` becomes a `CommandRedirect`: trimmed, with every run of whitespace collapsed to a single space. `'  migration \t apply '` is stored as `'migration apply'`.

Normalizing rather than rejecting, because a command path is a sequence of segments and whitespace is only the separator between them. A doubled space has exactly one sensible reading, and nobody writing `from` means a literal tab. Making the author's obvious intent a build failure would be the wrong trade — and normalizing covers tabs for free, which a hand-written empty-segment check would have missed.

One case is still refused: a `from` that is empty once normalized, since nothing can produce it.

## What testing turned up

The mounted-path side had no test at all, and writing one showed its protection works differently than it looks. `buildCommandTree` validates group declarations first, so a malformed multi-segment mount path fails there as an unknown group (`references unknown group 'migration '`), and `insertCommand`'s own empty-segment rule effectively only fires for the empty path.

Those four tests stay, pinning behaviour that had never been covered. They are not otherwise related to this change.


wmadden (approved):

wmadden:
@coderabbitai review

wmadden:
@coderabbitai review

PR #153, opened 2026-08-11
The engine renders: an aligned card, aligned tables, drawn trees, a palette, and the width to fit them in

wmadden-electric:
## What this looks like

`prisma-cli auth workspace list`, before and after:

```
name  id  status              name      id    status
Acme Inc  ws_1  current  →    Acme Inc  ws_1  current
Globex  ws_2                  Globex    ws_2
```

`prisma-cli auth whoami`, whose card is the shape 36 commands use:

```
status: signed in                status:     signed in
user: bob@example.com       →    user:       bob@example.com
workspace: Acme Inc              workspace:  Acme Inc
```

And the tree the style guide has specified since it was written, which until now no block could express:

```
✘ contract
├─ ✘ table user
│  └─ ✔ primary key: id
└─ ✔ table post
```

Headers and keys cyan, connectors dim, glyph and label coloured by the node's status. None of that was reachable before: `makeUi` emitted two escape sequences, bold and dim, and nothing else in the engine emitted colour at all.

## What was broken

The engine shipped a common renderer that did almost no rendering. `renderBlock` was `columns.join("  ")` — no sizing, no padding. The key/value card printed `label: value`, having lost the aligned, accent-coloured column the commander shell drew. The tree emitted two-space indents and had zero users. There was no way to draw a migration graph's lane gutter at all, and `OutputStream` was `{ write(text) }`, so the engine could not learn the terminal's width even from the bin.

Both consumers had already lost rendering to it. The platform port flattened `auth workspace list`, `project list` and `agent status`, recorded as accepted divergences. The ORM's renderers ship colourless, and two of them call `createColors({ useColor: true })` to force colour past `NO_COLOR` because there is no other way to get any.

## The decision

The engine renders; a command describes. So the fix is to the renderer, not to give commands a canvas.

- **`table` aligns** and **`fields` draws the card again** — keys padded to one column, toned, with the dim `│` rail available as `fields.rail` for the commands that want it.
- **`tree` draws connectors and status glyphs**, matching the style guide verbatim.
- **`drawing` is added** as the escape hatch for layout the engine cannot derive: the migration DAG's lane gutter, where lane assignment comes from a BFS over the graph and one hue must reach the gutter cell, the node glyph and the label alike.
- **`Text = string | Span[]`** wherever a block takes display text. Spans carry tone, never escape sequences, so the engine measures on text and colour cannot break alignment. The pad-versus-colour bug that exists in shipped code becomes unrepresentable.
- **`Status` is separate from `Tone`.** One says what happened and picks the glyph; the other says what colour to paint. They were fused in `summary.tone`, which meant a tree node could not be a failure painted in its branch lane's colour.
- **`ui.width`** is stderr's width, unbounded when stderr is not a terminal. Only the command knows what to sacrifice, so it is told the room it has; if it overruns anyway the engine prints it unmodified.

Two corrections ride along. Colour resolution keyed off stdout while blocks render to stderr, so `cmd > file` disabled colour for output a human was watching — it now follows the stream that is actually printed to. And every one of the CLI's four failure surfaces drew `✖`, while the style guide and `docs/product/output-conventions.md` have both specified `✘` all along; all four now draw `✘`.

## Nothing here was invented

The palette is not a set of tastes. Every tone is the colour a shipping Prisma CLI already uses for that meaning, checked at source during review:

| | |
| --- | --- |
| `heading`, and the eight other platform semantics | `packages/cli/src/shell/ui.ts` |
| `identifier`, `ref`, `highlight` | the ORM CLI's formatters in prisma/prisma |
| `color-1` … `color-6` | the ORM's migration-graph `LANE_COLORIZERS`, in order |

The indexed colours are six rather than the eight first sketched, because six is what the ORM's lane rotation is, and it excludes red deliberately so no lane can read as an error. Basic ANSI has no two further hues that stay distinguishable from those six; reaching eight would mean a 256-colour palette neither CLI uses. `colorette` and `string-width` are the colour and measurement libraries both repos already depend on.

## How to believe the card came back

`packages/cli/tests/v8-golden-rendering.test.ts` renders a `fields` block through the engine and asserts it equals what the commander shell's own `renderFieldRows` produces. It is a comparison against the original code in another package, building its own colour instance and sharing nothing with the engine — not a byte string someone typed into a test.

Alignment is proven not to depend on colour: a table whose cells carry spans is byte-identical to the same table in plain text once escapes are stripped, which a colour-first-then-pad implementation would fail. A CJK cell is measured in terminal columns rather than code units.

## What this does not do

No command binds `ui` yet — all 54 platform presenters are still written `human: () => [...]`, so the engine colours what it draws itself and nothing more. Spans, `drawing` and `ui.width` get their first callers when each family converts its renderers, which is the next step for both. Next actions and diagnostics still render as plain strings, so half of stderr is coloured and half is not; that is the natural follow-on slice and it is recorded as such.

The rail is available but no command sets it. Alignment and colour return to all 36 cards at once; which of them want the rail is a per-command judgement the engine cannot make.

## Alternatives considered

**A separate `graphic` presenter alongside `human`** — rejected: a drawing is human output, so it belongs in the human channel's grammar, and a parallel presenter would need its own channel and colour handling.

**Routing rich renderings through `Presentations.stdout`** — rejected: that is the machine channel, and a drawing's consumer is a human.

**Letting commands draw tables and trees themselves** — rejected: if every command reimplements every form of rendering on an open canvas, the common renderer has no purpose.

**Keeping `NO_COLOR` absolute over an explicit `--color`** — rejected: a flag on the invocation beats a standing environment default. `NO_COLOR` still beats the terminal.

---

Supersedes #148, which carried this spec alone. The spec is in this PR at `.drive/projects/prisma-cli-v8/specs/engine-colour.md`, with the four amendments the implementation survey produced recorded in its §8.


wmadden:
@coderabbitai review this

wmadden (approved):

PR #165, opened 2026-08-12
Attach the tarballs before the Release publishes, because published means frozen

wmadden-electric:
## The grounding example

The first real `next` publish (run 31618278670) shipped both packages to npm and then failed its final step:

```text
https://github.com/prisma/prisma-cli/releases/tag/v8.0.0-rc.1
HTTP 422: Cannot upload assets to an immutable release.
##[error]Process completed with exit code 1
```

`v8.0.0-rc.1` exists, marked pre-release — with no tarballs attached, and none can ever be attached now.

## The decision

This repo's releases are immutable: publishing freezes assets and tag. So the Release step now creates the Release as a **draft with the smoked tarballs already attached**, then publishes it through the API by the draft's id — assets first, publish second. A draft's tag does not exist yet, so `gh release edit <tag>` cannot address it; the id lookup handles that. A rerun that finds the Release already published logs that there is nothing to repair and succeeds, instead of failing a run whose npm publish completed.

`v8.0.0-rc.1` itself stays assetless — immutability cuts both ways. Its tarballs remain retrievable as the run's workflow artifacts and from npm itself. The next release gets its assets.

## Alternatives considered

- **Upload then publish via `gh release create` without `--draft`** — what the code did; gh creates the release published and uploads after, which immutability rejects.
- **Disable immutable releases** — the setting is a supply-chain protection worth keeping; the pipeline should conform to it, not the reverse.
- **Delete and recreate v8.0.0-rc.1 with assets** — immutable releases cannot be deleted; and even if they could, replacing a published release quietly would defeat the immutability guarantee.


wmadden (approved):

wmadden-electric:
Superseded by #166 — same two commits on a branch cut from current main. This PR's diff showed the entire already-merged S7 slice because #164 was squash-merged, which orphaned this branch's original commits.

PR #170, opened 2026-08-12
Plan one working happy path for every command the binary answers to

wmadden-electric:
**The `prisma` binary answers to 89 commands. 40 have an e2e test. 49 do not.**

Counted by walking `--help` from the root through every group, diffed against the `describeCommand` blocks in `packages/cli/e2e/*.e2e.ts`. Nothing in the suite names a command the binary does not mount.

## What the test has to be

One happy path per command, run through the assembled binary, doing the thing and succeeding. Not `--help`, not "it parses its arguments", not "it fails honestly without credentials" — those prove a command is reachable, not that it works.

The reason this repo has to do it rather than lean on the owning repos: **the product repos test their commands exhaustively, but none of them can reproduce the configuration this binary assembles.** The mount paths, the engine version actually linked, several families' config sections at once, the credential wiring, and the dependency tree inside the published tarball exist only here. A command can pass its owner's entire suite and still be broken the moment it is mounted.

26 of the 49 come from other packages — 22 from `@prisma/orm-toolchain`, 4 from `@prisma/composer` — and not one is run by this repo's e2e suite.

## The fixtures decide the order, not the command groups

Most of the 49 are blocked on the same few things, so the work is to build those and the commands follow in bulk:

- **A deployed service unlocks ~16** — the seven `service deployment` verbs, `service open`, `service show`, the five `service domain` verbs, and `build logs`. Highest value and most expensive: it creates real infrastructure the suite must then remove.
- **A project directory with a contract unlocks most of the ORM family** — `format`, `orm init`, `contract`, the eight `migration` verbs, `ref`. No API needed; `tests/fixtures/orm-project` already exists to build on.
- **A database unlocks the rest** — `db init|schema|sign|update|verify` and `migrate`. The suite already creates Prisma Postgres databases, so this is wiring.
- Smaller ones: a GitHub repo for `git connect|disconnect`, a second workspace for `project transfer`, a project with real backups for `postgres restore`.

## Four need a ruling before anyone writes them

- **`auth login`** drives a browser, and every other test authenticates with a service token — so the one command whose job is obtaining a credential is the one the suite cannot exercise normally.
- **`feedback`** posts real feedback to the CLI team unless it can be pointed at a test endpoint.
- **`composer dev`/`log`** never settle on their own, so "happy path" needs a definition — started, served one request, shut down cleanly.
- **`lsp`** speaks LSP on stdio; its happy path is an answered `initialize`, a different harness from everything else here.

Plan only, no code. Nothing to be written as a skipped test in the meantime — a skipped test reads as coverage.

### DIFF
--- a/.drive/projects/prisma-cli-v8/plan.md
+++ b/.drive/projects/prisma-cli-v8/plan.md
@@ -195,6 +195,42 @@ Recorded so they are not lost between slices.
   authenticated socket. Shelved, not rejected — unlike `service deploy`,
   which is not coming back in that shape.
 
+## E2E coverage of every command the binary answers to
+
+Spec R7 requires product-repo e2e in prisma/prisma and composer. It says nothing about this repo covering the grammar it ships, and the gap is bigger than the platform commands alone.
+
+**The `prisma` binary answers to 89 commands. 40 have an e2e test. 49 do not.** Counted by walking `--help` from the root through every group, against the `describeCommand` blocks in `packages/cli/e2e/*.e2e.ts`; nothing in the suite names a command the binary does not mount.
+
+### What the test has to be
+
+**One happy path per command, run through the assembled binary, doing the thing and succeeding.** Not `--help`, not "it parses its arguments", not "it fails honestly without credentials". Those prove a command is reachable; they do not prove it works.
+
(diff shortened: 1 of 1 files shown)


wmadden (approved):

wmadden-electric:
Closing as superseded, and stale enough to be worth saying why.

The convention this planned already exists in enforced code — `packages/cli/tests/e2e-coverage.test.ts` requires a real-API happy path for every mounted command, holds the exclusions and the backlog with reasons, and fails CI when an entry rots. #178 extended it with eight new happy paths and corrected two of its claims against the API.

This document went stale in a day, which is the argument against keeping it:

- It says 40 of 89 commands are covered. It is 48 now.
- It says a deployed service "unlocks about 16" and that only `composer deploy` can produce one. #178 showed this repo can produce one through the management API, and that the domain verbs then need a DNS record on top — the most important fact in the section, and the one it gets wrong.
- Its "four that need a ruling" are all already in `EXCLUSIONS` with reasons.

The one paragraph worth keeping is the rationale — why this repo owes a happy path per command, rather than leaning on the product repos, with #171 as the proof. That moves into the header of `e2e-coverage.test.ts`, next to the rule it justifies.

PR #180, opened 2026-08-13
ADR 0004: resources live on branches, a deploy produces a Version, and every surface speaks one language

wmadden-electric:
One config file declares the system; deploying it on a branch creates that branch's own copy of every declared resource:

```text
git branch: main                        git branch: pr-123
config declares service "web"           same file via git, plus new "worker"
        │  deploy                               │  deploy
        ▼                                       ▼
Branch: main                            Branch: pr-123
└── Service web                         ├── Service web        ← same key, so
    └── Version 14   (live)             │   └── Version 1        "the same service";
                                        │                        a different
                                        └── Service worker       platform resource
                                            └── Version 1
```

That picture is the whole proposal. This ADR makes it the official model for the three surfaces currently improvising their own answers — the unified CLI's grammar, the Console, and the platform API that Composer deploys against — and fixes the vocabulary that goes with it.

Built up in order, each section grounded in the diagram above:

1. **Branches are isolated environments, and resources live on them.** The two `web` rows are two platform resources. What makes them "the same service" is the shared config key, carried between branches by git — not a project-level registry, which this model deliberately doesn't have. The honest cost is stated: renaming a key is delete-and-create until an explicit rename annotation exists.
2. **A deploy produces a Version.** "Deployment" the noun retires; the control plane's own table is already named `ComputeVersion`, so this is the API catching up to its schema. One aliasless cut across API, Console, CLI, fields, and error codes. The old `latestDeploymentId` splits into two honestly named facts: newest version and live version.
3. **The evaluated config graph becomes a branch-level Topology record**, and drift (people changing resources directly — always allowed) becomes something the platform can *show* rather than something anyone pretends to prevent.
4. **Every run that changes a branch writes a Build Run** — a purely informative, workspace-level audit record. The platform never executes user code, so the record names its reporter and links to the CI run; laptop deploys write one too, so every Version has provenance.

Two names are reserved for the future so nobody designs something unrelated under them: **Database Version** and **subgraph deployment**.

Alternatives we weighed and rejected — project-level identity with per-branch "instances", keeping "Deployment", deriving the topology from deploy state, "Build Job" — are at the end of the ADR, with reasons.

**Review ask:** this is Status: Proposed. The decision being requested is the model itself — the four numbered points above — not the prose. If you disagree with one, the Alternatives section is where your objection may already be addressed.

### DIFF
--- a/docs/architecture/adrs/0004-platform-domain-model-and-language.md
+++ b/docs/architecture/adrs/0004-platform-domain-model-and-language.md
@@ -1,0 +1,281 @@
+# ADR 0004 - Resources live on branches, a deploy produces a Version, and every surface speaks one language
+
+## At a glance
+
+One config file declares the system. Deploying it on a branch creates that
+branch's own copy of every declared resource:
+
+```text
+git branch: main                        git branch: pr-123
+config declares service "web"           same file via git, plus new "worker"
+        │  deploy                               │  deploy
+        ▼                                       ▼
+Branch: main                            Branch: pr-123
+└── Service web                         ├── Service web        ← same key, so
+    └── Version 14   (live)             │   └── Version 1        "the same service";
+                                        │                        a different
+                                        └── Service worker       platform resource
+                                            └── Version 1
+```
+
+Everything in this ADR is visible in that picture. The two `web` rows are
+two platform resources — branches are isolated environments — and what makes
+them the same service is the shared key `web`, declared in one config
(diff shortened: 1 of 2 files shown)


wmadden-electric:
Withdrawn: this ADR belongs in pdp-control-plane (the domain model is the control plane's), and its grounding claim was wrong — the application is declared in code per Composer ADR-0003/0017 and PDP ADR-009, not by a config file. Superseded by the pdp-control-plane ADR.

PR #176, opened 2026-08-13
service logs keeps its name; brief for the platform team's HTTP endpoint

wmadden-electric:
Records only.

- `deferred.md`'s `service logs` entry gains the operator's naming ruling (2026-08-13): the command mounts as **`service logs`**, the legacy spelling — not under the `deployment` subgroup — and the CLI-side sequencing (build on contract confirmation; final wiring after the endpoint reaches the generated SDK).
- New `assets/briefs/deployment-logs-http-endpoint.md`: the self-contained ask for the platform/control-plane team — serve the existing log/terminal records as NDJSON over authenticated GET with `cursor` resume, grounded in their own repo (`streamLogs.ts` is a 1 s poll relay, so HTTP loses nothing; `streamBuildLogs.ts` is the in-repo template) with links to the CLI project's design docs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #181, opened 2026-08-13
service logs returns: page reads over plain HTTP, --follow polls the cursor

wmadden-electric:
```
$ prisma service logs --help
USAGE
  prisma service logs [--service name] [--project id-or-name] [--deployment id]
                      [--tail n] [--from-start] [--follow] [<service>]

Read logs for a deployment of the service

Examples:
  prisma service logs
  prisma service logs --tail 500
  prisma service logs --follow
  prisma service logs --deployment dep_123 --from-start
```

**Do not merge before prisma/pdp-control-plane#4886 deploys** — this command reads the plain-GET page mode that PR adds; merged early it ships a command that 404s in production.

**The decision.** `service logs` was the last unported command in the CLI — shelved in the S2c port because the deployment-logs endpoint spoke only WebSocket and the credential rules forbid a command building its own authenticated socket. The API team's page-read contract (#4886) removed the need for any engine transport work: one authenticated `GET` returns one NDJSON page and closes with a terminal record carrying a resume cursor. This PR restores the shelved command on that contract, under its legacy name (operator ruling).

**What shipped.**
- **Page mode by default**: the last 100 lines (`--tail n`, or `--from-start`), then exit 0 — the `kubectl logs` shape. The legacy command followed by default; that divergence is recorded.
- **`--follow`**: re-requests with each terminal record's cursor on a poll interval. Polling, not push — the WebSocket live tail stays the platform's, and the engine's socket transport design stays shelved for the live-streaming date.
- **Target resolution ported verbatim from the shelved S2c handler** (explicit `--deployment` resolved globally then checked against the project; otherwise the service's live deployment), with its error shapes. Only the transport died.
- **Refusals instead of silence**: `--follow` with no cursor to resume from errors (`SERVICE.LOGS_NO_CURSOR`) rather than silently re-reading the same tail forever (the reverted mutation ran 20s into the test timeout — that loop was real); a body that ends without its terminal record errors (`SERVICE.LOGS_INCOMPLETE`) in both modes rather than exiting 0 on a partial page. Printed lines still print; what's refused is claiming completeness.
- `build logs`' NDJSON reader extracted to `lib/ndjson.ts` and shared — duplicating chunk-boundary handling is the drift class that produced the S8 workspace-filter defect.

Records: the slice contract is `.drive/projects/prisma-cli-v8/specs/service-logs.md` (including the retired SDK-risk premise — the pinned SDK's operation type already carried `tail`/`from_start`/`cursor`); divergences in `assets/s2/parity-divergences-service-logs.md`; `deferred.md`'s logs entry closes. This branch also carries #176's records (the naming ruling and the platform-side brief) — if #176 merges first, this rebases clean.

**Alternatives.** Waiting for the platform SDK bump: unnecessary, the premise was wrong. A follow that quietly re-tails when it loses its cursor: rejected as indistinguishable from progress. One error code for both no-cursor and truncation: rejected because the two facts differ (a complete page follow can't continue vs an incomplete page both modes must refuse) and their next-actions differ.

### DIFF
--- a/.drive/projects/prisma-cli-v8/assets/briefs/deployment-logs-http-endpoint.md
+++ b/.drive/projects/prisma-cli-v8/assets/briefs/deployment-logs-http-endpoint.md
@@ -1,0 +1,113 @@
+# Brief: serve deployment logs over HTTP (for the platform/control-plane team)
+
+Written 2026-08-13, for an agent working in `pdp-control-plane` with
+no prior context on the CLI project. Operator: Will Madden.
+
+## The ask, in one paragraph
+
+Add an HTTP variant of the deployment-logs endpoint: the same
+log/terminal records `GET /v1/deployments/{deploymentId}/logs`
+streams over its WebSocket today, served instead as newline-delimited
+JSON over a plain authenticated `GET`, keeping the `cursor` resume
+semantics and the in-band terminal record. This was agreed in
+principle on 2026-08-12 (team answer, via Will): **"HTTP instead of
+WebSockets would be acceptable, as long as we can add live streaming
+at a later date."** This brief is the concrete version of that ask.
+
+## Why the CLI needs it
+
+The v8 Prisma CLI (repo `prisma/prisma-cli`, built on
+`@prisma/cli-engine`) must ship a `service deployment logs` command —
+the last unported command in its entire surface. The old
+implementation reached around the engine: it took a raw token and let
+`@prisma/compute-sdk`'s `streamLogs` build a `wss:` URL and set the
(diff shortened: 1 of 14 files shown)


wmadden (approved):

wmadden-electric:
The merge blocker is cleared: prisma/pdp-control-plane#4886 deployed to production today. `GET /v1/deployments/{deploymentId}/logs` serves the paged NDJSON mode live (verified via https://api.prisma.io/v1/doc — 200 `application/x-ndjson` alongside the 101 upgrade), and the SDK regeneration was dispatched by the same deploy.

PR #187, opened 2026-08-17
Engine 0.1.1: the vendor table is built in, not loaded from ci-info at runtime

wmadden-electric:
## The failure this fixes

Found by composer's engine-0.1.0 adoption (prisma/composer#234). Under Bun, evaluating any composer config died with `CONFIG.EVALUATION_FAILED ... vendors.map is not a function`.

## The root problem, not the symptom

The engine's published artifact loaded another package's **internal file at runtime**: `import vendors from "ci-info/vendors.json" with { type: "json" }`. That file is CJS-owned — ci-info's own `index.js` requires it — so in any process that loads both (composer's config graph reaches ci-info through orm-toolchain's config-loader), one file lives in two module systems. Bun hands the later CJS require the `{__esModule, default}` wrapper instead of the array. Changing the engine's access mode (an earlier revision used `createRequire`) would only shuffle that hazard; removing the runtime reach removes it.

## The decision

The table is **inlined at build time** (tsdown `noExternal` on the one JSON file). The built engine touches nothing of ci-info's at runtime; `ci-info` moves to devDependencies. The design intent survives intact: the engine ships built, so "upgrade ci-info and rebuild" — which every release already does — is still what teaches it a new provider, and ci-info's `isCI` remains unusable by design (it reads real `process.env` at import; the engine only reads host-injected env).

Pinned by `no-esm-json-import-in-dist.test.ts` (no vendors.json reference survives in the output), and the engine's import-purity conformance run is back to fully strict — no allowances. Verified under real Bun 1.3.13 and Node: engine dist + CJS ci-info coexist; the pre-fix combination reproduced composer's failure verbatim.

## A second problem we found and did NOT fix here

While tracing the failure we found a dependency that should not exist: `@prisma/composer-prisma-cloud` — a library that applications install — imports `@prisma/orm-toolchain/config-loader`, which belongs to the ORM's command-line tooling. ADR 0004 says libraries must not depend on CLI packages, and this import is how an engine internal was able to break composer's users at all. Fixing it means deciding where `config-loader` should actually live, which is Will's decision, so this PR leaves it alone. Recorded in `.drive/projects/prisma-cli-v8/deferred.md` with the full import chain and the options.

## The bump

`pnpm bump-cli-engine-version patch` (#186's first real use) → engine `0.1.1`, shell pin, lockfile in one commit; the shell's pin-exception triples re-keyed, reopened by the version move as designed. On merge the next publish run ships `0.1.1`; composer's #234 then updates its pinned engine version and its remaining red jobs clear.

## Alternatives considered

**`createRequire` of the same file** — shipped in this PR's first revision, rejected on review: it still reaches into ci-info's internals at runtime, and matches only the current owner's access mode. **Hand-copy the table** — loses the upgrade property for no gain over build-time inlining. **Use ci-info's `isCI`** — reads the real environment at import, forbidden by design. **Wait for Bun** — the registry behaviour is arguably Bun's bug, but hosts run today's Bun, and the engine artifact should not be exposed to any runtime's dual-registry semantics in the first place.

### DIFF
--- a/packages/cli-engine/src/ci.ts
+++ b/packages/cli-engine/src/ci.ts
@@ -1,4 +1,4 @@
-import { createRequire } from "node:module";
+import vendors from "ci-info/vendors.json" with { type: "json" };
 import type { Runtime } from "./runtime";
 
 type Env = Readonly<Record<string, string | undefined>>;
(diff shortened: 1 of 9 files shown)


wmadden (approved):

packages/cli-engine/tests/no-esm-json-import-in-dist.test.ts
```diff
@@ -0,0 +1,41 @@
...
+ * The vendor table is read through the CJS require cache, never as an
+ * ES-module JSON import. Under Bun 1.3, a JSON file that enters the ES
+ * module registry poisons a later CJS `require` of the same file — it
+ * returns the `{__esModule, default}` wrapper instead of the array — and
+ * ci-info's own index.js requires its vendors.json. The engine cannot
+ * control what else a host process loads, so its built output must use
+ * the same cache ci-info does. Found by composer's engine-0.1.0
+ * adoption (prisma/composer#234): `vendors.map is not a function` when
+ * any config graph reached `import { isCI } from "ci-info"` under bun.
+ */
```
wmadden, line 5: Are these comments still accurate? Do the need to be so long?

packages/cli-engine/src/ci.ts
```diff
@@ -18,12 +18,25 @@ interface Vendor {
...
+ * Inlining is a correctness requirement, not an optimization. The file
+ * is CJS-owned — ci-info's own index.js requires it — and a published
+ * artifact that loads someone else's internal file at runtime puts one
+ * file in two module systems in the same process. Under Bun, the
+ * ES-module registry entry poisons the later CJS require (the array
+ * arrives as its `{__esModule, default}` wrapper), which broke config
+ * evaluation for every composer-under-bun host the moment the config
+ * graph also loaded ci-info (prisma/composer#234). With the table
+ * embedded, the built engine touches nothing of ci-info's at runtime
+ * and ci-info is not a runtime dependency at all. Pinned by
```
wmadden: Do we really need 50 line comments?

packages/cli-engine/tests/no-esm-json-import-in-dist.test.ts
wmadden-electric: Trimmed to four lines — accurate for the inlined approach (the earlier ten described the superseded createRequire revision).

packages/cli-engine/src/ci.ts
wmadden-electric: Cut to seven lines: why isCI is unusable, why the table is inlined, and the pinning test. The narrative lives in the PR body and commit message where it belongs.

PR #200, opened 2026-08-18
Engine 0.2.0: the changed engine must ship, and CI now forces the bump

wmadden-electric:
## What

`npx prisma@next --version` crashes with `SyntaxError: The requested module '@prisma/cli-engine' does not provide an export named 'readActiveAccessToken'`.

**Why:** #183 added engine exports and #184 changed the engine further, and neither bumped the engine version. The registry's `@prisma/cli-engine@0.1.1` was published ten minutes before #183 merged, and published versions are immutable — so `prisma@8.0.0-rc.4` imports names the engine it installs does not have. CI could not catch it: the workspace builds against its local engine, which has the exports.

## Changes

- `pnpm bump-cli-engine-version minor` → engine **0.2.0** (the delegated spawn protocol changed in #184, so this is the pre-1.0 breaking bump). The next publish run on `main` ships it.
- **New CI job** (`engine-version` in test.yml, backed by `scripts/check-engine-version.mjs` + tests): a PR that changes `packages/cli-engine/` while leaving its version at one the registry already has fails, with the bump command in the error. This is the check whose absence let rc.4 ship broken.

## What this does not fix on its own

Both products declare the engine as an **exact** peer at 0.1.1, and npm fails with `ERESOLVE` on the mismatch (verified empirically). The repair sequence is:

1. Merge this → engine 0.2.0 publishes.
2. `prisma/composer` and `prisma/prisma` re-declare the peer at 0.2.0 and release (`composer-cli` 0.7.1, `orm-toolchain` 8.0.0-rc.3).
3. This repo picks up those versions and cuts `8.0.0-rc.5` — the first version where `npx prisma@next` works again.

### DIFF
--- a/.drive/projects/prisma-cli-v8/deferred.md
+++ b/.drive/projects/prisma-cli-v8/deferred.md
@@ -429,6 +429,16 @@ CLI does not do, and each restarts as engine work if wanted:
   for exactly this kind of consumer, or cut the cloud extension's use
   of it. Spans both product repos; Will decides which.
 
+## The rc.4 engine mismatch (2026-08-18) — the repair chain
+
+`prisma@8.0.0-rc.4` on `next` crashes on import: #183 and #184 changed the engine without bumping its version, so the registry's `@prisma/cli-engine@0.1.1` (published ten minutes before #183 merged) lacks exports the CLI imports. The registry is immutable and both products peer the engine exactly, so npm fails with ERESOLVE on any mismatch; the repair is a chain, in order:
+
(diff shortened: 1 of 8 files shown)


wmadden (approved):

wmadden-electric:
The first push failed PR Quality on this repo's own conformance check — correctly: with the engine at 0.2.0 and both products peering 0.1.1, the sandbox install resolves two engines. That mismatch is inherent to any engine transition (a family cannot peer an engine version that is not on the registry yet), so 5ea5d9c adds the two exception entries the conformance machinery was designed with for exactly this case: keyed on the exact 0.1.1/0.2.0 pair, printed in every run as "allowed for now", and removed by the 8.0.0-rc.5 bump PR. While they stand, a premature release could ship the two-engine install they describe — that is the argued cost of the transition, and why the rc.5 PR must remove them. Dev-channel conformance now reports 0 failing, 3 allowed.

PR #214, opened 2026-08-19
fix(auth): identify workspace sessions by account

AmanVarshney01:
## Why

Local workspace sessions can belong to different Prisma accounts. The CLI previously showed workspace names without account identity, making an old personal login indistinguishable from the work account signed into Console.

## Changes

- Show account identity in workspace lists, selection prompts, login, workspace use, and workspace logout. Prefer email, then name, then id.
- Add nullable user metadata to JSON results and mark workspace lists as local-sessions, not the complete Console membership list.
- Fetch workspace name and account metadata together through one /v1/me request. Remove the separate workspace lookup; no API change is needed.
- Cache missing legacy-session names and account metadata without replacing known fields. Ordinary session reads and logout remain local-only.
- Keep metadata lookup best-effort with a three-second timeout. Failed lookups leave sessions usable and fall back to token claims.
- Recheck credentials under the state lock before saving metadata. Concurrent logout cannot resurrect a session; a replacement login cannot inherit the previous account's metadata.
- Preserve account metadata during token rotation. Keep token material out of command output.

The shared cli-engine contract, dependencies, and CI workflows are unchanged. Session identity stays CLI-local, because `@prisma/composer-cli` and `@prisma/orm-toolchain` pin the engine at exactly 0.4.0 and an engine bump fails the conformance check until both republish. The CLI reads the CLI-local fields through typed guards.

- `enrichSessions()` sends no request that cannot succeed. It skips a session with an expired access token, and it skips the user lookup for a token that belongs to no user.
- The plain stdout rows of `auth workspace list` keep the columns they have on main (workspace, id, status). The user appears in the table and in JSON only.
- `auth workspace list` offers `auth login` as a next action only when there are no sessions, as on main. The label is now "Authorize a workspace".
- `auth workspace use <workspace>` resolves an explicit argument against local state and makes no request. It looks metadata up only when the argument matches no stored session, so a session saved before names were stored still resolves by name. The interactive picker still looks metadata up.

## Review cleanup

Reviewed all 16 PR files and merged current main, including the OAuth-denial fix. Addressed the single-request review and the legacy workspace-name persistence finding. Removed the unnecessary logout lookup, duplicate comment, test-only wrapper, and redundant metadata-filter type assertion. Existing output and security tests remain; focused regressions cover name-only enrichment and concurrent replacement/removal.

## Verification

On 15fe7f5b (local run): full CLI suite 990 passed with 2 platform-dependent skips; root typecheck, lint, and `check:error-reference` passed. New command-level tests cover the lookup path: `auth workspace list` calls it, `auth workspace use <workspace>` does not when the argument matches, and does once when it matches nothing. Conformance, E2E, and the manual checks below were not repeated on this commit.

On efe5ab22 (local run): full CLI suite 987 passed with 2 platform-dependent skips; build, root typecheck, lint, release-channel conformance, and `scripts/check-engine-version.mjs` passed. E2E and the manual checks below were not repeated on this commit.

On fbb3dfd3:

- Full CLI suite: 985 passed, 2 platform-dependent skips.
- Build, root typecheck, and lint passed.
- Local E2E: 6 passed, 48 credential-dependent tests skipped because no E2E service token is configured locally.
- Built CLI against a local HTTP API: distinct accounts, legacy workspace-name persistence, cached metadata, workspace selection, local logout for cached and uncached sessions, and no tokens in JSON all passed. Two initial sessions made exactly two /v1/me requests; subsequent list/use/logout commands made none.
- The prior single-request change was also checked against a stalled HTTP server: login saved the session and returned after approximately three seconds.

No real browser login was repeated and no real local credentials were modified. Fresh CI checks must complete before merging.

## Scope

This lists workspace sessions authorized on this machine, not all remote workspace memberships. It does not fix account switching on the OAuth browser page. create-prisma can consume items[].user to identify accounts in its deployment picker.


packages/cli/src/auth/credential-manager.ts
```diff
@@ -207,26 +312,37 @@ export class FileCredentialManager implements CredentialManager {
       this.#actAs({ kind: "session", workspaceId });
     }
 
-    const name = await this.#lookUpWorkspaceName(credential, workspaceId);
-    if (name === undefined) return created;
+    const [name, identity] = await Promise.all([
+      this.#lookUpWorkspaceName(credential, workspaceId),
+      this.#lookUpSessionIdentity(credential, workspaceId),
+    ]);
+    if (name === undefined && identity === undefined) return created;
```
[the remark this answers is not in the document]
AmanVarshney01: Fixed. Login now re-reads the session under the state lock after both metadata lookups, including the no-metadata path. It returns the replacement session, or the existing session-ended error if the record was removed. The existing concurrency tests now assert the returned identity and both metadata/no-metadata outcomes. Verified 106 targeted tests, 974 CLI tests (2 skipped), typecheck, and lint.

packages/cli/src/auth/session-metadata.ts
wmadden, line 1: Does this mean we need to do 2 new network requests just to show metadata? Would it be better to augment one of the existing API responses to carry the data we need?

packages/cli/src/commands/auth/workspace-list.ts
```diff
@@ -71,9 +69,28 @@ function listPresentations(result: WorkspaceListResult): Presentations {
...
+          sessionLabel(session),
+          session.workspaceId,
+          session.workspaceId === result.selectedWorkspaceId ? "current" : "",
+        ]
+          .join("  ")
+          .trimEnd(),
+      ),
     json: () => serializeWorkspaceList(result),
-    next: () => (result.sessions.length === 0 ? [LOGIN_NEXT_ACTION] : []),
+    next: () => [
```
wmadden-electric, line 84: Only-in-the-CLI behaviour: this emits a next action even when the list has rows. Every other list command emits none, or only emits one when the list is empty. Please restore the `result.sessions.length === 0` condition and keep the new "Authorize a workspace" label for that case.

wmadden-electric (dismissed):
The single-request question is settled. `/v1/me` returns `user` and `workspace` together, and this PR deletes the separate `/v1/workspaces/{id}` lookup, so login makes the same number of requests as before. Thanks for reworking it that way.

One change requested, then this looks good to merge.

## The next action on a populated list

Every other list command in the CLI either emits no next action at all (`branch list`, `bucket list`, `bucket key-list`, `postgres list`, `postgres backup-list`, `postgres connection-list`) or emits one only when the list is empty (`project env-list`, and `auth workspace list` on main). This change would make `auth workspace list` the only one that appends a suggestion to a result that already has rows.

A next action is for the dead end, when the operator has nothing to act on. A populated list is not a dead end — they got what they asked for, and `auth workspace use` is discoverable from help.

Please restore the `sessions.length === 0` condition, and drop the sentence in `docs/product/output-conventions.md` saying the command always offers `auth login`. Keep the new label: "Authorize a workspace" describes the model better than "Sign in", because each login authorizes one workspace.

## Non-blocking, for later

`auth workspace use <workspace>` now calls `sessionsForDisplay`, so it can make a `/v1/me` request before doing a purely local switch. Failed lookups are not recorded, so on a stalled connection the three-second timeout is paid on every run until one lookup succeeds. Plain offline is cheap, because the connection fails immediately — the cost lands on captive portals, VPN blackholes and packet loss.

This is small in practice. Sessions created by this version store both fields at login, so in steady state there are no requests at all, and the legacy sessions that do trigger a lookup clear themselves the first time one succeeds.

Worth scoping later rather than now: `workspace use` with an explicit argument does not need the lookup to resolve the reference, only to fill one output row. `workspace list` and the interactive picker are where the identity is the point. The same reasoning you already applied to keep `workspace logout` local-only.

## Smaller points, none blocking

- `AUTH.WORKSPACE_AMBIGUOUS` meta carries both `workspaceIds` and `sessions[].workspaceId`. The second supersedes the first. Worth picking one while the protocol is still pre-release.
- The command-level tests never run the real `enrichSessions` path, because the test helper patches `sessions()` instead. `enrichSessions` has good direct unit coverage, so this is only a gap in proving that `auth workspace list` triggers a lookup at all.

The rest reads well. The duck-typed `sessionsForDisplay` is a fair workaround for the pinned engine contract, a replacement login cannot inherit the previous account's metadata, rotation preserves `user`, and no token material reaches any output.

packages/cli/src/commands/auth/workspace-list.ts
AmanVarshney01: Fixed in 15fe7f5b. The `result.sessions.length === 0` condition is back, so a populated list emits no next action. The empty case keeps the "Authorize a workspace" label. I also dropped the "always offers `auth login`" sentence from `docs/product/output-conventions.md` and updated the command test and both golden expectations.

AmanVarshney01:
@wmadden-electric the blocking change is in 15fe7f5b (replied on the thread). On the non-blocking points:

**`auth workspace use <workspace>` and the lookup.** Done in the same commit. An explicit argument now resolves against local state and makes no request, so a stalled connection no longer costs the three-second timeout on a local switch. One exception: when the argument matches no stored session, the command looks metadata up once before failing. Without that, `use <name>` would stop resolving a session saved before names were stored. The interactive picker still looks metadata up, because identity is the point there.

**Command-level coverage of the lookup.** Added three tests in `auth.test.ts`: `auth workspace list` calls `enrichSessions` and renders what it returns; `auth workspace use <workspace>` never calls it when the argument matches; and calls it once when the argument matches nothing.

**`workspaceIds` and `sessions[].workspaceId` on `AUTH.WORKSPACE_AMBIGUOUS`.** Left both for now. `workspaceIds` is already on main and in `docs/reference/error-reference.md`, which syncs to docs.prisma.io, so I would rather not remove a published field inside this PR. I added the missing `sessions` meta to the error reference. Happy to drop `workspaceIds` in a follow-up if you want the protocol narrowed before release.

AmanVarshney01:
@coderabbitai review

wmadden (approved):

PR #221, opened 2026-08-24
Bump to version 8.0.0-rc.9: fix the prisma package's product pins

wmadden-electric:
prisma@8.0.0-rc.8 crashes on every invocation: the wrapper package's own dependency list still pinned @prisma/orm-toolchain@8.0.0-rc.4 and @prisma/composer-cli@0.11.0, so the published bin resolved the old ORM family keys and the mount table's lookups came back undefined ("Cannot read properties of undefined (reading 'needs')"). Only packages/cli's pins were bumped in #218; the conformance sandbox masked the divergence by hoisting the good version from @prisma/cli's manifest, which a real `npm install prisma` does not do.

This aligns packages/prisma's pins with packages/cli (composer-cli 0.12.0, orm-toolchain 8.0.0-rc.5) and bumps to 8.0.0-rc.9. The check gaps (no pin-equality check across the two manifests; the tarball bin-start not exercising a prisma-only install) are recorded in the deferred ledger.


wmadden:
@coderabbitai review

PR #228, opened 2026-08-25
Remove the deprecated defineConfig alias before launch

wmadden-electric:
The engine's deprecated `defineConfig` re-export of `definePrismaConfig` was the only `@deprecated` surface in the repository, and nothing consumes it: `prisma/config` re-exports `definePrismaConfig` by name, and the composer/ORM families' same-named helpers are their own, not imports of the engine's. Removed ahead of launch, riding the in-flight unpublished engine 0.2.3 so no additional release chain is needed. Tests and doc prose follow.


wmadden-electric:
Folded into the config-file-resolution PR per operator instruction; this branch is superseded.

PR #229, opened 2026-08-25
fix(cli): restore Deno support for packaged CLI

AmanVarshney01:
## Summary

Restore Deno startup for the packaged Prisma CLI without copying or changing credential-storage behavior.

- keep `@prisma/credentials-store` as the single auth-storage implementation
- bundle it into both published CLI binaries, so Deno does not resolve its broken transitive conditional export at runtime
- move it from a published runtime dependency to a build-time dependency

## The failure

With Deno 2.9.5, the current published RC fails before parsing any command:

```console
$ deno run --minimum-dependency-age=0 -A npm:prisma@8.0.0-rc.9 --version
error: Cannot find module ".../xdg-app-paths/8.3.0/src/mod.deno.ts"
```

The same failure occurs through `npm:@prisma/cli@8.0.0-rc.9`.

The startup dependency chain is:

```text
prisma / @prisma/cli
  -> FileTokenStorage
  -> @prisma/credentials-store
  -> xdg-app-paths
```

`xdg-app-paths@8.3.0` declares a Deno conditional export at `./src/mod.deno.ts`, but that file and its imports are absent from the npm tarball. Because auth storage is statically imported, every CLI command fails, including `--version` and `orm init`.

Upgrading `@prisma/credentials-store` does not fix this: its latest release still uses `xdg-app-paths@8.3.0`.

## Why bundle the store?

The CLI should not duplicate credential persistence. `@prisma/credentials-store` remains the only implementation used by `FileTokenStorage`; this PR changes packaging only.

Both published binaries already use bundled entry points. Adding the store to `deps.alwaysBundle` makes the build resolve its Node-compatible implementation once and include it in the shipped JavaScript. The published package therefore does not ask Deno to resolve `xdg-app-paths`, while Node and Deno execute the exact same auth code.

This avoids vendoring storage logic, patching an installed package, lazy-importing auth, or adding a Deno-specific runtime branch. No auth-file schema, path, locking behavior, command, flag, or output changes.

## Verification

- `pnpm typecheck`
- `pnpm lint`
- `pnpm --filter @prisma/cli test` — 61 files, 958 passed, 1 skipped
- `pnpm --filter @repo/cli-conformance test` — 5 files, 66 passed
- `PUBLISH_CHANNEL=dev pnpm check:conformance` — 0 failing findings
- packed `prisma@8.0.0-rc.10`, installed it into a clean npm sandbox, and ran under official Deno 2.9.5:

  ```console
  prisma orm init --target postgres --authoring psl --skip-install --yes --format json
  ```

  It completed successfully and wrote the expected config, contract, database client, TypeScript config, environment example, package manifest, and Git metadata files.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -52,7 +52,6 @@
     "@prisma/cli-engine": "workspace:0.2.3",
     "@prisma/composer-cli": "0.14.0",
     "@prisma/compute-sdk": "0.39.0",
-    "@prisma/credentials-store": "^7.8.0",
     "@prisma/management-api-sdk": "1.55.0",
     "@prisma/orm-toolchain": "8.0.0-rc.7",
     "@vercel/detect-agent": "^1.2.3",
@@ -63,6 +62,7 @@
   },
   "devDependencies": {
     "@prisma/composer": "0.14.0",
+    "@prisma/credentials-store": "^7.8.0",
     "@repo/cli-conformance": "workspace:8.0.0-rc.10",
     "@repo/cli-telemetry": "workspace:8.0.0-rc.10",
     "@repo/tsconfig": "workspace:8.0.0-rc.10",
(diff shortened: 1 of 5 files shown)


wmadden (approved):
Thanks @AmanVarshney01 👍🏻

PR #244, opened 2026-08-27
fix: include local Postgres runtime in Prisma CLI

AmanVarshney01:
## Summary

- declare `@prisma/dev` as a runtime dependency of both published Prisma CLI packages
- keep the consolidated `prisma` wrapper manifest identical to `@prisma/cli`
- teach import-purity and packed-tarball conformance about the runtime Composer resolves dynamically

## Why

Composer local development deliberately resolves `@prisma/dev` through the Prisma CLI installed by the generated application so the application controls its Prisma version. The published consolidated `prisma` package did not carry that runtime, which made local Postgres fail in strict workspace layouts unless templates added `@prisma/dev` themselves.

The generated application should depend on `prisma`; it should not need to know about this internal local-emulator package.

## Validation

- `pnpm --filter @prisma/cli test` — 961 passed, 1 skipped
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:conformance` — packed and checked all five release subjects
- installed the packed `prisma` tarball in a strict pnpm Turborepo and verified Composer resolves its nested `@prisma/dev@0.25.2`
- with the companion Composer resolution fix, ran the generated app through build, migrations, local emulators, and an HTTP request returning seeded data

## Companion fix

- prisma/composer#268 resolves through the exported Prisma package manifest.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -52,6 +52,7 @@
     "@prisma/cli-engine": "workspace:0.3.0",
     "@prisma/composer-cli": "0.16.0",
     "@prisma/compute-sdk": "0.42.0",
+    "@prisma/dev": "0.25.2",
     "@prisma/management-api-sdk": "1.69.0",
     "@prisma/orm-toolchain": "8.0.0-rc.8",
     "@vercel/detect-agent": "^1.2.3",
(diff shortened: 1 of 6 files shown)


AmanVarshney01:
Closing this for now. I opened it before confirming the upstream PR plan with the team.

PR #245, opened 2026-08-28
Ship the prisma-platform-core-concepts skill in the prisma package

tylerhogarth:
The `prisma` package now ships an agent skill defining the Prisma Platform's core concepts, the same way `@prisma/composer` and the ORM anchor packages ship theirs. Consumers receive it through `prisma skills sync`.

## Changes

1. **Skill**: `skills/prisma-platform-core-concepts/SKILL.md` covers structures, hierarchies, relationships, and workflows: the workspace/project/branch model, branches as preview environments, the two deploy paths (GitHub app recommended, CLI `deploy --stage`), services and versions, the Compute sleep model, Prisma Postgres, object storage, the env-var class/scope model, and the local development stack. It is deliberately not a CLI reference; it routes command surfaces to `--help`. `skills/README.md` records the authoring rules.
2. **Packaging**: `scripts/stage-skills.mjs` stages skills into the package at `prepack`, selected by each skill's `metadata.library` frontmatter. `packages/prisma` adds `skills` to `files`; the staged copy is gitignored.
3. **Version stamping**: `scripts/set-version.ts` now stamps `metadata.library_version` in every skill on a lockstep bump, via the new `scripts/skill-frontmatter.ts` helpers.
4. **Verification**: `scripts/check-skill-packaging.mjs` (root script `check:skill-packaging`, new `pr-quality` job) packs the package the way publish does and checks the tarball carries the skill, stamped with the packed version and byte-identical to the tracked source.
5. **Allowlist**: `prisma` joins `SKILL_SOURCE_PACKAGES` in `packages/cli/src/lib/skills/allowlist.ts`, the deliberate per-package trust decision that lets sync install from it.

## Why

Staging is frontmatter-driven rather than a hardcoded table so the skill list has one source of truth, and it runs at `prepack` rather than build so a stale turbo cache can never produce a tarball without it. The tarball check exists because none of the failure modes (missing `files` entry, failed staging, hand-edited stamp) show up in unit tests; they show up in what npm uploads. The pattern is a direct port of the one proven in prisma/composer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #247, opened 2026-09-01
Rewrite CLI help as a self-documenting manual

tylerhogarth:
Makes `--help` the reference for the CLI: every card teaches the product model it depends on, written for a reader (human or agent) who has never used Prisma and has no skill installed.

## Changes

- **Help standard**: New `docs/product/cli-help-standard.md` defines the wording contract and its rationale: rows stand alone, group briefs lead with scope, group cards define the terms their rows use, no internal vocabulary (binding, resolved, pinned). `cli-style-guide.md` gains the card rules and `AGENTS.md` binds future commands to the standard.
- **Engine**: `HelpSpec` gains a `WorkflowStep` type and group/root help a `workflow` field; the renderer (`packages/cli-engine/src/execution/help.ts`) draws a `Workflow` section of `$`-prefixed, two-column steps before Examples. Covered by new tests in `execution.test.ts`.
- **All repo-owned help text**: root card, 20 group cards, and 58 leaf commands rewritten. Group briefs follow "short lead, then scope" ("Manage S3-compatible object-store buckets for a project. CRUD operations and access keys"); descriptions define concepts at first use (Workspace > Project > Branch > services/databases/buckets, linking, service versions, env scopes); flag briefs state when to use them and spell defaults in plain language. Workflows added to root, project, postgres, bucket, service, and service domain.

## Why

Help ships with the binary and always matches the installed version, so it is the one interface every consumer is guaranteed to have. Writing it for an agent that has never heard of Prisma removes the churn of undefined vocabulary: one card read, one correct invocation, no doc round trip and no reliance on an installed skill. Behaviour claims in the new text were verified against handlers before writing; one-time secret semantics, consent tokens, and deploy-on-push descriptions match the code.

The "Prisma Next" strings and `deploy`/`dev` summaries ship in `@prisma/orm-toolchain` / `@prisma/composer-cli` and need PRs in those repos.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,6 +31,10 @@ Unit tests may still mock, and should — error paths and edge cases belong ther
 
 Why this rule exists: `prisma project list` reported "No projects found." and exited 0 for a workspace holding 15 projects, and every project-scoped command was broken with it. The unit suite covered that command thoroughly and passed throughout, because its fixtures supplied both sides of every comparison — the credential's workspace id and the API's were the same hand-written string, while the real API returns a `wksp_` prefix that the credential does not carry. A test that writes both sides of a comparison can only confirm what its author already believed. Mocks are still the right tool for error paths and edge cases; they cannot tell you what the API actually returns.
 
+## Help Text
+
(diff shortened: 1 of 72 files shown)


tylerhogarth:
Addressed the five help-text comments in 68b4533 (service-token override on both workspace cards, key role default, env update --file mode, link's create path, live-version claims softened to "at most one"). The three SKILL.md comments target a file that is no longer in this PR: the review ran against the pre-force-push head that still contained #245's commits. They apply to #245.

PR #257, opened 2026-09-11
feat(cli-engine): send deploy source and client headers to the Management API

gregory-boch-prisma:
## Problem

The Management API records Compute deploys in its analytics, but it cannot tell which tool sent a request or whether it ran in GitHub Actions. The CLI's requests carry only the runtime's default User-Agent (`node` or `Bun/x.y.z`), which any Node or Bun script also sends.

## Change

Every request from the engine's Management API client (`ctx.api`) now carries three headers:

| Header | Value |
| --- | --- |
| `x-prisma-client-name` | `prisma-cli` |
| `x-prisma-client-version` | the CLI version, from `getCliVersion()` |
| `x-prisma-deploy-source` | `github-action` when `GITHUB_ACTIONS` is `"true"`, otherwise `cli` |

- The headers are added by an openapi-fetch `onRequest` middleware on `sdk.client`, the same way the SDK adds `Authorization`.
- The version reaches the engine through a new optional `cliVersion` field on `ManagementApiClientConfig`. A host that does not set it simply omits `x-prisma-client-version`.
- `ctx.api` carries Composer's build reports during `prisma deploy` and the `prisma service` commands, so both are covered.
- `@prisma/cli-engine` is bumped from 0.3.0 to 0.3.1 with `pnpm bump-cli-engine-version patch`, because published engine versions cannot change.

The headers change analytics only. The API behaves the same with or without them, and they carry no user data: just the tool name, its version, and whether the run is in GitHub Actions.

The server side is prisma/pdp-control-plane#5284. A matching change for Composer's alchemy client covers the deployment create and start calls, which do not go through `ctx.api`.

## Verification

- `pnpm typecheck` and `pnpm lint` are clean.
- `pnpm --filter @prisma/cli-engine test`: 830 passed. New tests check the headers inside and outside GitHub Actions, and that `x-prisma-client-version` is left out when `cliVersion` is not set. They go through the real client with a stubbed `fetch`.
- `pnpm --filter @prisma/cli test`: 964 passed, 2 skipped. `bin.test.ts` now expects `cliVersion` in the assembled config.
- `pnpm --filter @prisma/compute test`: 11 passed.
- No command was added or changed, so the e2e suite is not affected.

### DIFF
--- a/packages/cli-engine/package.json
+++ b/packages/cli-engine/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prisma/cli-engine",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "description": "The execution engine of the unified Prisma CLI.",
   "type": "module",
   "exports": {
--- a/packages/cli-engine/src/execution/api-client.ts
+++ b/packages/cli-engine/src/execution/api-client.ts
@@ -152,7 +152,30 @@ async function constructClient(
     authBaseUrl: config.authBaseUrl,
     tokenStorage: storage,
   });
+  sdk.client.use(
+    deploySourceMiddleware(invocation.runtime.env, config.cliVersion),
+  );
   return { client: sdk.client, pinned: { active, storage } };
+}
+
+/** Analytics-only headers: the Management API records them on deploy events and behaves the same without them. */
+function deploySourceMiddleware(
+  env: Record<string, string | undefined>,
+  cliVersion: string | undefined,
+) {
+  return {
+    onRequest({ request }: { request: Request }): Request {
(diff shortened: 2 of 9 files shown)


gregory-boch-prisma:
Parking this PR. Bumping `@prisma/cli-engine` here changes the exact engine version that Composer and the ORM toolchain pin, so it can't ship on its own. The deploy headers will go out with the next planned engine release instead. Until then, the deploy source is covered by prisma/composer#290 (Composer deploys) and prisma/pdp-control-plane#5285 (GitHub Actions OIDC deploys).

PR #274, opened 2026-09-17
feat(engine): the loader tells a config file where it is

wmadden-electric:
## At a glance

```ts
// prisma.config.ts, unchanged for users
export default definePrismaConfig({
  orm: ormConfig({ contract: './contract.prisma', migrations: { dir: './migrations' } }),
});
```

While the engine's loader evaluates this file it publishes the file's directory through an `AsyncLocalStorage` kept on `globalThis` under `Symbol.for('prisma.config.baseDir')`. `ormConfig` reads it as it runs and resolves its own paths, so the section a command receives already holds absolute paths and the directory they were resolved against:

```ts
// what the orm command family receives, wherever the command ran from
{ baseDir: '/app', contract: { output: '/app/contract.json', ... }, migrations: { dir: '/app/migrations' } }
```

## The decision

A relative path inside `prisma.config.ts` is relative to that file. The loader is the one party that knows which file it is evaluating, so it tells the file, and the family's config helper, the one party that knows which of its fields are paths, resolves them on the spot. The engine never learns what a section contains, and config authors write nothing new. The full design, the layering constraint, and the rejected alternatives are in [ADR 253](https://github.com/prisma/orm/blob/main/docs/architecture%20docs/adrs/ADR%20253%20-%20Config%20paths%20resolve%20against%20the%20file%20that%20wrote%20them.md) in prisma/orm.

## The bug this fixes

The engine handed sections over exactly as written and nothing told a command family which file they came from. The ORM's commands resolved their paths against `ctx.cwd`. Verified with `prisma` 8.0.0-rc.13 and `@prisma/orm-toolchain` 8.0.0-rc.8:

```
exp/
  sub/
    prisma.config.ts   # contract: './contract.prisma'
    contract.prisma
```

From `exp/sub`, `prisma contract emit --config ./prisma.config.ts` writes `exp/sub/contract.json`. From `exp`, `prisma contract emit --config ./sub/prisma.config.ts` fails with `CONTRACT.SOURCE_LOAD_FAILED` looking for `exp/contract.prisma`.

## What changes

- **`withBaseDir(dir, evaluate)`** runs `evaluate` inside an `AsyncLocalStorage` scope carrying `dir`; **`baseDir()`** reads it. Both are exported, with the key as `BASE_DIR_KEY`, for any other loader or helper; a family package can also reach the store without importing the engine, since the key is `Symbol.for`. It is an `AsyncLocalStorage` rather than a plain value so evaluations that overlap in one process, such as a language server loading several projects, each see their own directory. Only `withBaseDir` creates the store; `baseDir()` reads it if present, so a family helper needs neither the engine nor `node:async_hooks` to read it and runs under any runtime.
- **The loader** wraps each chain file's evaluation in `withBaseDir(dirname(path), ...)`. Since #233 the loader discovers and evaluates the chain itself, one file at a time, so every file on the chain sees its own directory; a parent config's `./migrations` resolves under the parent, a child's under the child, before the two are merged.
- **Engine version.** #233 already moved the engine to 0.5.0, which is not on the registry yet, so this change ships under it; `check-engine-version` passes.

Nothing changes in the command context, `definePrismaConfig`, or any user's config file.

## Tests

- `withBaseDir`: publishes the directory during the evaluation and clears it after; restores the outer directory after a nested evaluation; keeps two overlapping evaluations apart; clears it when the evaluation throws.
- `loadConfig`: a fixture whose section reads the store as it runs comes back with the fixture's directory, both discovered and named with a relative `--config` from another cwd; on a two-file discovery chain each file records its own directory; the store is empty once the files have been read.
- The main export surface test lists the three new exports.

`pnpm --filter @prisma/cli-engine test`: 38 files, 937 tests, rebased on #233. Repository typecheck and lint pass.

## Consumer side

[prisma/orm#30328](https://github.com/prisma/orm/pull/30328) makes `ormConfig` read the slot and resolve its paths, has the ORM's own loader publish the slot the same way, and refuses a section that reaches a command without `baseDir`. It merges independently; once this engine ships, the unified CLI resolves the ORM's paths correctly.

## Status and release order

An engine version bump trips the conformance check until the mounted product packages peer on 0.5.0, and they cannot until 0.5.0 is published. The sequencing of that window (ADR 0004) is an operator call.

**This engine must ship before the ORM toolchain that depends on it.** The ORM's validator refuses a section without `baseDir`, so an `@prisma/orm-toolchain` release carrying prisma/orm#30328 mounted by a shell still on engine 0.4.0 would fail every ORM command for every user.


wmadden-electric:
Closing in favour of a schema-driven design: `defineConfigSection({ name, schema })` with a `path` keyword, the engine loader validating and resolving path fields per chain file from the declaration. No ambient base directory, no per-family resolution code. A fresh PR follows.

